PR #12, opened 2017-03-09
add AirDrop support

vab9:
This commit allows the app to receive and handle audio files sent through AirDrop and other file-sharing means.

* make this app a file handler for various audio file-types in Info.plist
* add a function to handle incoming files in AppDelegate.swift
    * this function simply moves the file from Documents/Inbox/ to Documents/
* ignore the Inbox/ directory when loading audiobooks in ListBooksViewController.swift
    * (the Inbox/ folder is created for incoming files by the OS)
* add the filename as a fallback option for the audiobook title if the metadata can't be parsed properly

### DIFF
--- a/Audiobook Player/AppDelegate.swift
+++ b/Audiobook Player/AppDelegate.swift
@@ -30,6 +30,29 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
         UserDefaults.standard.set(nil, forKey: "sleep_timer")
         return true
     }
+    
+    func application(_ app: UIApplication,
+                              open url: URL,
+                              options: [UIApplicationOpenURLOptionsKey : Any] = [:]) -> Bool {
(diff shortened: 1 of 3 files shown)


GianniCarlo:
Awesome work @vab9 !, there's only one case not covered. When the app is already running in background and I import a book via airdrop, I have to manually refresh the list to see the new book in the list.

One solution is to notify `ListBooksViewController` (once the file was moved from the folder `inbox`) via the Notification center and register it to call directly the function `loadFiles`. That should cover that.

Thanks for simplifying `loadFiles` and adding the `alternateTitle` which is much more user friendly 🎉

vab9:
Great! Glad to hear it works. I'm looking forward for the new version.

GianniCarlo:
I'll be adding the commit covering the case I mentioned. Thanks again @vab9 for this awesome pull request 🎉

PR #18, opened 2017-09-17
add higher speed options

e7mac:
Since audiobooks have a high variety of reading speeds, it'd be great to have more speed options.


GianniCarlo:
Sure!, thanks for the suggestion

PR #21, opened 2017-10-27
Feature: Smart Rewind

bryanrezende:
Added a feature that I found surprisingly useful in another audiobook app that I use, "Bound", where if you leave a book paused for longer than 10 minutes it will rewind by a bit to let you remember and get a sense of where you were. 

Made some minor changes to how we access userDefaults in the PlayerManager according to what I understand to be best practices, as well as adding a constants file that we can use to enable better localization in the future by collecting all strings there. 

What do you think?

### DIFF
--- a/Audiobook Player/AppDelegate.swift
+++ b/Audiobook Player/AppDelegate.swift
@@ -22,11 +22,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
         Fabric.with([Crashlytics.self])
         let defaults:UserDefaults = UserDefaults.standard
         
-        //If app has never been launched...
-        if defaults.bool(forKey: UserDefaultsConstants.isFirstLaunch) {
-            //Set smartRewindEnabled and isFirstLaunch to false
+        // Perfrom first launch setup
+        if !defaults.bool(forKey: UserDefaultsConstants.completedFirstLaunch) {
+            // Set default settings
             defaults.set(true, forKey: UserDefaultsConstants.smartRewindEnabled)
-            defaults.set(false, forKey: UserDefaultsConstants.isFirstLaunch)
+            
+            defaults.set(true, forKey: UserDefaultsConstants.completedFirstLaunch)
         }
 
         UIApplication.shared.statusBarStyle = .lightContent
--- a/Audiobook Player/AppDelegate.swift
+++ b/Audiobook Player/AppDelegate.swift
@@ -20,6 +20,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         // Override point for customization after application launch.
         Fabric.with([Crashlytics.self])
+        let defaults:UserDefaults = UserDefaults.standard
+        
+        //If app has never been launched...
+        if defaults.bool(forKey: UserDefaultsConstants.isFirstLaunch) {
+            //Set smartRewindEnabled and isFirstLaunch to false
+            defaults.set(true, forKey: UserDefaultsConstants.smartRewindEnabled)
+            defaults.set(false, forKey: UserDefaultsConstants.isFirstLaunch)
+            
+            //Ensure user defaults are saved immediately
+            defaults.synchronize()
+        }
 
         UIApplication.shared.statusBarStyle = .lightContent
         
--- a/Audiobook Player/AppDelegate.swift
+++ b/Audiobook Player/AppDelegate.swift
@@ -27,9 +27,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
             //Set smartRewindEnabled and isFirstLaunch to false
             defaults.set(true, forKey: UserDefaultsConstants.smartRewindEnabled)
             defaults.set(false, forKey: UserDefaultsConstants.isFirstLaunch)
-            
-            //Ensure user defaults are saved immediately
-            defaults.synchronize()
         }
 
         UIApplication.shared.statusBarStyle = .lightContent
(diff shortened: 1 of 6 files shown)


GianniCarlo:
hey @bryanrezende thanks so much for your contribution!, now that you mention it, I've experienced it too in another app and found it useful. I like the idea of having a constants file too.

I'll test it out later today, my only feedback so far is `defaults.synchronize()` which is no longer necessary: https://developer.apple.com/documentation/foundation/userdefaults/1414005-synchronize I'll leave a comment in the corresponding line

bryanrezende:
Awesome, I'm happy to help! I've just recently fallen in love with audiobooks and found that there's no option, free or paid, that has all of the options I'd like while still being good-looking and responsive. I figured I'd try my hand at making one and put it out for free for the community, then saw your project and loved it. I will probably be bugging you with a LOT of PRs in the coming weeks, if you don't mind :) 

And thanks for pointing out the synchronize thing, I had completely missed that! I'll get that changed and pushed up.

Audiobook Player/AppDelegate.swift
```diff
@@ -20,6 +20,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         // Override point for customization after application launch.
         Fabric.with([Crashlytics.self])
+        let defaults:UserDefaults = UserDefaults.standard
+        
+        //If app has never been launched...
+        if defaults.bool(forKey: UserDefaultsConstants.isFirstLaunch) {
```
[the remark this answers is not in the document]
GianniCarlo: maybe we can move this `if` to the bottom of the function before `return true` and do something like:
```
.
.
.
UserDefaults.standard.set(nil, forKey: "sleep_timer")

//here comes the new code
if defaults.bool(forKey: UserDefaultsConstants.completedFirstLaunch) {
  return true
}

//Set smartRewindEnabled and completedFirstLaunch to true
defaults.set(true, forKey: UserDefaultsConstants.smartRewindEnabled)
defaults.set(true, forKey: UserDefaultsConstants.completedFirstLaunch)

return true
}//end of function
```
it's not set in stone to call the constant `completedFirstLaunch`, I used that name in efforts to make clear what I mean 🤔. What do you think ?

Audiobook Player/PlayerManager.swift
```diff
@@ -250,6 +261,8 @@ extension PlayerManager: AVAudioPlayerDelegate {
             return
         }
         
+        defaults.set(self.identifier, forKey: UserDefaultsConstants.lastPlayedBook);
```
GianniCarlo: (This is just a comment) This looks interesting, are you planning on loading the bottom bar with the last played book on the next launch? I really like that idea 🤔

Audiobook Player/PlayerManager.swift
```diff
@@ -157,8 +158,18 @@ class PlayerManager: NSObject {
...
-                //get stored value for current time of book
-                let currentTime = UserDefaults.standard.integer(forKey: self.identifier)
+                //get stored value for current time of book in seconds
+                var lastPlayedPositionInSeconds = self.defaults.integer(forKey: self.identifier)
+                
+                
+                // If smartRewind is enabled and time since last play was 10 minutes (599s), rewind audiobook by 30 seconds or to start.
+                if let lastPlayTime:Date = self.defaults.object(forKey: UserDefaultsConstants.lastPauseTime+"_\(self.identifier)") as? Date, self.defaults.bool(forKey: UserDefaultsConstants.smartRewindEnabled)  {
+                    if Date().timeIntervalSince(lastPlayTime) > 599 {
+                        lastPlayedPositionInSeconds = max(lastPlayedPositionInSeconds - 30, 0)
```
GianniCarlo: just like we store the `lastPauseTime` on the pause event, I think we should clear the time once the smart rewind is triggered once when loading a book, otherwise we would risk having a side effect of triggering more than once for the same recorded time. What do you think 🤔?

Audiobook Player/AppDelegate.swift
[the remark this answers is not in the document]
bryanrezende: Good catch! I originally had the method reversed and then decided to "clean it up" grammatically, and ended up with bad logic. I like using "completedFirstLaunch" as the constant for clarity, but I personally like having a !completedFirstLaunch check and have all initial setup logic contained in there. It makes it a bit easier to read at a glance since it's all chunked in one spot and you can't miss the `return true` and be confused when reading it. This way we can always add any fresh "first run" logic and always know where it's all contained. I'm thinking about changing it to:

```
//If app has never been launched...
        if !defaults.bool(forKey: UserDefaultsConstants.completedFirstLaunch) {
            //Set smartRewindEnabled and isFirstLaunch to false
            defaults.set(true, forKey: UserDefaultsConstants.smartRewindEnabled)
            defaults.set(false, forKey: UserDefaultsConstants.completedFirstLaunch)
            
            //Ensure user defaults are saved immediately
            defaults.synchronize()
        }
```

What do you think?

Audiobook Player/PlayerManager.swift
bryanrezende: Hmmm that's probably a good idea. I can't help but think there's probably a cleaner way I can implement the smart rewind check that would get around this issue without having to null out the user default every time, but I can't think of it! I'll add this in!

Audiobook Player/PlayerManager.swift
bryanrezende: That's the plan! I just forgot to remove this line from the commit since it didn't play into the feature. There's a lot we can do if we have the last listened to book - I was also thinking of adding an opt-in setting to autoplay last book on launch.

Audiobook Player/AppDelegate.swift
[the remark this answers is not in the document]
GianniCarlo: makes sense what you say about `return true` 👌, let's go with your suggestion, but don't forget in that block of code to remove `defaults.synchronize` and to set to `true` `completedFirstLaunch` 👍

Audiobook Player/PlayerManager.swift
GianniCarlo: 👌 we can figure it out as we move along with the project :D

PR #22, opened 2017-10-28
Improvement: PlayerView UI Shows Cover Art

bryanrezende:
Modified the player view to show the cover art of whatever book is being listened to, centered above the play controls. 

In order to make sure the cover art never clashes with whatever the background color is, added a gradient where the top is the average color of all colors in the cover and it fades to black so that we always have good contrast for the controls. If for whatever reason this fails, we default to the original blue, and if we cannot find the album art, we use the app icon (we can make a new empty state if we want something original). 

Also changed the "three-dot" icon to base its color on whatever tint we provide so that we can use Chameleon's high contrast extension based on the average color. 

I set up my last branch a bit funky and then branched off of it, so I've got some extra code changes here (the smart rewind feature), but I figure instead of going through trying to properly undo the storyboard shenanigans, I'd push this up with some screenshots and get your input so I can keep making any changes. Going forward I'll make sure my master is clean and any features are on their own branches!

Here's some example screenshots, one with a dark book cover and one light: 

v2_1

v2_2

And with a possible drop shadow around the cover: 

v2_3

### DIFF
--- a/Audiobook Player/Base.lproj/Main.storyboard
+++ b/Audiobook Player/Base.lproj/Main.storyboard
@@ -366,7 +366,7 @@
                         <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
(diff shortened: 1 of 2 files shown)


GianniCarlo:
Awesome work @bryanrezende ! I had been thinking these past few days that I needed to do something about the blue in the app because lately it seemed to be too much of it now that the player is presented modally instead of pushed. The gradient for the player completely solves that problem, and I really love how the drop shadow around the cover kind of merges with the gradient, it makes it subtle yet effective 👍

GianniCarlo:
Before merging this pull request, I was checking how it looked with the app icon as the default cover art. I'm okay with it, let me know if this is what you had in mind @bryanrezende

bryanrezende:
I think it looks alright with the app icon as the backup image! Hopefully that’s something most people shouldn’t run into, but in the event they do nothing looks jarring. I think we can add a follow-up issue/ticket to make a new asset for an empty state in the player and list that could look more aesthetic.

PR #29, opened 2017-12-25
Sorting book list by title.

gpambrozio:
Sorting was broken for me, at least when using the newest OS.


GianniCarlo:
👍 thanks for the fix!

PR #30, opened 2017-12-25
Adding a way to boost the volume of the audio.

gpambrozio:
I feel that sometimes my volume is very low and I could use a boost. Not sure if this is desirable for everyone but since I have it already here it goes. Feel free to just reject it.


GianniCarlo:
Thanks @gpambrozio !, I'll test it out throughout the day 👍. While it's true that some may not want this, I think it's very valuable to make it available as an optional feature.

PR #39, opened 2018-03-16
Add global speed control

e7mac:
While listening to an audiobook, I do not want to set speed for every item manually. With this switch, users are able to set a global speed that will be used if the switch is on.


GianniCarlo:
Thanks for your contribution @e7mac ! I'll review it and try it out first thing tomorrow 👌

GianniCarlo:
Awesome! thanks @e7mac

Issue #49, opened 2018-03-29
Extend existing guards in PlayerViewController


pichfl:
Instead of adding a separate `if`, pull the query into the guard.

https://github.com/GianniCarlo/Audiobook-Player/blob/44ddec4c9359984acd7d66c568dc1190213ef037/Audiobook%20Player/PlayerViewController.swift#L217-L230

https://github.com/GianniCarlo/Audiobook-Player/blob/44ddec4c9359984acd7d66c568dc1190213ef037/Audiobook%20Player/PlayerViewController.swift#L232-L258

pichfl:
After talking to a colleague of mine I think we should keep this as is.

GianniCarlo:
I would argue the opposite in those specific cases, for the following reason: there's nothing happening after the `if let` statement, so if we already have a guard in place, I would rather do all the unwrapping there and then thave the other statements without any identantion. This idea is grabbed from the 'golden path' which states that we should check first for all the error states, so we can always have the success state at the end

pichfl:
You're right. I think I misunderstood mostly because of the `PlayerManager.sharedInstance.stop()` in the second block, which does not have to be there but come further down as well.

GianniCarlo:
closed with #50

PR #50, opened 2018-03-29
Refactor: Guards

pichfl:
This was caused by discussion I had with a colleague of mine regarding #49. I'm not completely confident I did this right, please check.


Audiobook Player/PlayerViewController.swift
```diff
@@ -202,35 +208,28 @@ class PlayerViewController: UIViewController {
     }
 
     @IBAction func didSelectChapter(_ segue: UIStoryboardSegue) {
-
         guard PlayerManager.sharedInstance.isLoaded() else {
```
GianniCarlo: let's use this guard to unwrap the optionals for the viewController and currentChapter

pichfl:
Updated the PR to match your comments

Issue #51, opened 2018-03-29
Remove DeckTransition dependency


pichfl:
Before beginning work on the player, I thought it might be good to clean up a bit. The new design relies on a more native look, which allows us to skip the dependency and use more regular presentation

GianniCarlo:
I've been considering this these past few days. Go for it 👍

GianniCarlo:
done in #55

PR #55, opened 2018-03-29
Feature: Remove DeckTransition

pichfl:
- Replace all custom segues
- Present as modal
- Add navigation controllers
- Add navigation bars and toolbar
- Limit to portrait
- Use default status bar color

Re #51


pichfl:
Pushed a squashed commit which should remove those last two instances.

Issue #44, opened 2018-03-27
New layout for Player 


pichfl:
- [x] Enable Gestures
- [x] Fix colliding gestures by limiting the axis
- [x] Update forward/rewind icons with versions without shadows, color from artwork
- [x] Allow for text inside forward/rewind icons to be adjustable

This also includes:

- [x] Refactor PlayerViewController
- [x] Simplify play speed controls
- [x] Simplify More menu
- [x] Abstract artwork and player controls

### Player Paused / Player Playing
<img src="https://user-images.githubusercontent.com/194641/37992801-063b88ee-320d-11e8-9fad-671c7569b7bf.png" alt="player paused" width="300"> <img src="https://user-images.githubusercontent.com/194641/37992799-06059a5e-320d-11e8-95ae-221e7e682ae5.png" alt="player" width="300">

### Speed controls / Chapter selection
<img src="https://user-images.githubusercontent.com/194641/37992800-062277a0-320d-11e8-8634-665b4ce16286.png" alt="player playspeed" width="300">
### Smaller phones
<img src="https://user-images.githubusercontent.com/194641/37992805-068f627a-320d-11e8-99de-88efb80157e0.png" alt="iphone 8 - player" width="300"> <img src="https://user-images.githubusercontent.com/194641/37992804-0673e8f6-320d-11e8-845b-ab3d1f4077f9.png" alt="iphone se - player" width="256">

(Click to preview in original size)

pichfl:
Closed in favor of PR #75

PR #57, opened 2018-03-31
Refactor: Extract Sleep timer

pichfl:
First part of the refactoring effort that builds up to the new layout.

Re #44


pichfl:
Do not merge yet, this has a few kinks left, ie. when you turn off the time the label does not reset (which it did in a previous iteration, but I tried to be too clever and sourced the original label text from the current label, which is obviously wrong)

pichfl:
Found and fixed my problem, regarding the label, feel free to throw more comments at me when doing a review 🙂

Audiobook Player/Player/SleepTimer.swift
```diff
@@ -0,0 +1,138 @@
...
+typealias SleepTimerCompletion = () -> Void
+typealias SleepTimerStart = () -> Void
+typealias SleepTimerProgress = (Double) -> Void
+
+final class SleepTimer {
+    static let shared = SleepTimer()
+
+    let durationFormatter = DateComponentsFormatter()
+
+    var timer: Timer!
```
GianniCarlo: what do you think about having this var as either:
`lazy var timer = Timer()`
or
`var timer: Timer?`
it would remove the `nil` checks and make the code look more 'swifty'(?). Let me know what you think

GianniCarlo:
Tried out the app on this branch, I found two things:
- When turning off the timer, playback stops
- If I set the timer, dismiss the player, open it again, the timer label doesn't update itself
Regarding this last point, wouldn't it be weird to have two timers in the screen at once? (playback time and sleep timer)

GianniCarlo:
there might be an `IBAction` not connected well for `setSleepTimer` in `PlayerViewController` 🤔 I had to connect it to get the timer button to work, otherwise it was crashing

Audiobook Player/Player/SleepTimer.swift
pichfl: Still learning swift, sounds good to me.

pichfl:
Pretty sure the disconnect happened between one of the refactoring steps. Will fix. 

Playback stopping when disabling the timer was unintentional, will investigate. 

Regarding the label: yes, makes sense, I forgot to set the initial state correctly. 

Regarding the two timers: we can put it back into the message, I just liked that you can see what’s going on right away. I don’t fear that people will mistake either of the fields as they have distinct positions and colors in the redesign.

pichfl:
I should have addressed all the issues with the first iteration of this PR. The time left is back in the message of the action sheet to keep this iteration closer to what is in the current version. I left the option to change the label in as well but commented it out.

Audiobook Player/Player/PlayerViewController.swift
```diff
@@ -220,128 +209,44 @@ class PlayerViewController: UIViewController {
...
-
-        alert.addAction(UIAlertAction(title: "In 5 Minutes", style: .default, handler: { _ in
-            self.sleep(in: 300)
-        }))
-
-        alert.addAction(UIAlertAction(title: "In 10 Minutes", style: .default, handler: { _ in
-            self.sleep(in: 600)
-        }))
+    @IBAction func setSleepTimer() {
+        let actions = SleepTimer.shared.actionSheet(
```
GianniCarlo: This should be called something along the lines of `actionSheet` since `actions` is not an array of actions, but an instance of `UIAlertController`

Audiobook Player/Player/SleepTimer.swift
```diff
@@ -0,0 +1,118 @@
...
+        timeLeft = seconds
+        timer = Timer.scheduledTimer(timeInterval: 1.0, target: self, selector: #selector(update), userInfo: nil, repeats: true)
+
+        RunLoop.main.add(self.timer!, forMode: RunLoopMode.commonModes)
+    }
+
+    private func reset() {
+        alert.message = defaultMessage
+        
+        guard timer != nil else {
```
GianniCarlo: this shouldn't be necessary, that line after this: `timer?.invalidate()` should only work if timer is != nil, otherwise it won't do anything

GianniCarlo:
👍 Awesome!

PR #61, opened 2018-04-04
Refactor: Simplify PlayerManager and attached classes

pichfl:


Audiobook Player/Player/MoreOptions.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Audiobook Player
+//
+//  Created by Florian Pichler on 01.04.18.
+//  Copyright © 2018 Florian Pichler.
+//
+
+import UIKit
+
+class MoreOptions {
+    static let shared = MoreOptions()
```
GianniCarlo: I don't think `MoreOptions` should be a singleton, there's no state or variable that it keeps. We could make the function `actionSheet` a class function so it would be called like so: `MoreOptions.actionSheet(actions:...`

GianniCarlo:
Change base branch to `next/major`

Audiobook Player/Player/MoreOptions.swift
pichfl: 😅 yes, I went a little overboard there. I think I should be able to put this action sheets back into the PlayerViewController

pichfl:
Updated to match comments. Please see PR #57 first, as this PR here was based on that branch. Once you're happy with these changes, I'll squash the commits.

pichfl:
I tried to do a little too much at once. Replaced by #63 and future PRs.

PR #62, opened 2018-04-06
Refactor: Separate player view into subviews

pichfl:
There are a few open ends to tie up, but it hopefully shows the general direction.

- [x] Play/Pause button not updated in all cases


Audiobook Player/Extensions.swift
```diff
@@ -52,5 +52,6 @@ extension Notification.Name {
         public static let bookPaused = Notification.Name(rawValue: "com.tortugapower.audiobookplayer.book.pause")
         public static let bookEnd = Notification.Name(rawValue: "com.tortugapower.audiobookplayer.book.end")
         public static let bookChange = Notification.Name(rawValue: "com.tortugapower.audiobookplayer.book.change")
+        public static let bookPlayback = Notification.Name(rawValue: "com.tortugapower.audiobookplayer.book.playback")
```
GianniCarlo: When will this event be fired?

GianniCarlo:
progress bar for the book isn't being updated yet

pichfl:
Pushed a new version that updates the progress bar.

PR #63, opened 2018-04-07
Refactor player toolbar actions to action sheets

pichfl:
Both "Speed" and "More" can follow the same interaction as "Timer", making the interactions more consistent and the code much simpler.

This should only be merged after #57, I'll rebase as necessary.


GianniCarlo:
👌 no comments, only rebase is necessary

PR #67, opened 2018-04-12
The grand refactoring

pichfl:
Please do not merge yet, it's def. incomplete and I created several things that need to be fixed before this can be merged. Feel free to add comments anyway. 

I can't promise I'll be able to separate all the changes into smaller PRs so I'm not trying right now, but I can do so if there is no other way. This replaces #62


GianniCarlo:
so far it looks great! I'll delve a little bit deeper later and run it locally to see how it goes 👍

Audiobook Player/Extensions/Extensions.swift
```diff
@@ -52,5 +51,57 @@ extension Notification.Name {
         public static let bookPaused = Notification.Name(rawValue: "com.tortugapower.audiobookplayer.book.pause")
         public static let bookEnd = Notification.Name(rawValue: "com.tortugapower.audiobookplayer.book.end")
         public static let bookChange = Notification.Name(rawValue: "com.tortugapower.audiobookplayer.book.change")
+        public static let bookPlaying = Notification.Name(rawValue: "com.tortugapower.audiobookplayer.book.playback")
```
GianniCarlo: we need to delete `updateTimer` then

Audiobook Player/Models/Book.swift
```diff
@@ -0,0 +1,148 @@
...
+
+                UserDefaults.standard.set(nil, forKey: UserDefaultsConstants.lastPauseTime+"_\(self.identifier)")
+            }
+        }
+
+        return lastPlayedPosition
+    }
+
+    var percentCompleted: Int = 0
+    var percentCompletedString: String = "0%"
```
GianniCarlo: we could have this one be a computed property too based on `percentCompleted`

Audiobook Player/Models/Book.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    // MARK: Chapters
+
+    var chapters: [Chapter] = [] {
+        didSet {
+            self.updateCurrentChapter()
+        }
+    }
+    var currentChapter: Chapter?
+    private var cachedChapter: Chapter?
```
GianniCarlo: what's the use for `cachedChapter` in relation to `currentChapter`?

Audiobook Player/Models/Book.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    private func updateCurrentChapter() {
+        // Don't change anything if we have no chapter or are still in range of the current
+        if
```
GianniCarlo: I'd rather we use `guard` statements where there are only `return`s

GianniCarlo:
The pan gesture to dismiss the player should only be enabled when dragging from the top of the player, right now you can drag from anywhere in the player to dismiss it. This might be debatable actually 🤔, tell me what you whink

Audiobook Player/Models/Book.swift
pichfl: Whelp, that is an artifact from some step of my refactoring. No need for that, will clean up

pichfl:
I would argue that it is actually quite useful, especially on big devices. We do not have any other gesture that could interfere and running this on my personal phone feels quite natural.

Audiobook Player/Models/Book.swift
GianniCarlo: I didn't specify, but I'd rather make `percentCompletedString` a `get` only variable, that way we handle one less state in this class

GianniCarlo: That `get` would be based on `percentCompleted`

GianniCarlo:
noted, no problem about leaving the gesture as it is. I'd really like to merge this into `next/major`, let me know what other things you consider that are missing to complete this refactor (I don't expect the redesign of the player to be completed in this pr btw). I think this is a solid base for `3.0` and I'd really like to have this as a starting point for my next commits as soon as possible 👍

pichfl:
Will wrap up tonight 🙂

GianniCarlo:
Awesome!

PR #69, opened 2018-04-16
Autoplay Bugfixes

e7mac:
Some notifications are not explicitly called on the main thread so the UI sometimes does not update.
Further, there was an error with the logic that caused books to not autoplay in some cases.

### DIFF
--- a/Audiobook Player/PlayerManager.swift
+++ b/Audiobook Player/PlayerManager.swift
@@ -95,14 +95,18 @@ class PlayerManager: NSObject {
             do {
                 self.audioPlayer = try AVAudioPlayer(contentsOf: book.fileURL)
             } catch {
-                NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.errorLoadingBook, object: nil)
+                DispatchQueue.main.async {
+                    NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.errorLoadingBook, object: nil)
+                }
                 completion(nil)
                 return
             }
 
             guard let audioplayer = self.audioPlayer else {
                 //notify error
-                NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.errorLoadingBook, object: nil)
+                DispatchQueue.main.async {
+                    NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.errorLoadingBook, object: nil)
+                }
                 completion(nil)
                 return
             }
@@ -402,7 +406,6 @@ extension PlayerManager: AVAudioPlayerDelegate {
             if Int(audioplayer.currentTime) >= chapter.start {
                 self.currentChapter = chapter
                 let chapterString = "Chapter \(chapter.index) of \(self.chapterArray.count)"
-                //notify
                 let userInfo = ["chapterString": chapterString, "fileURL": self.currentBooks.first!.fileURL] as [String: Any]
 
                 //notify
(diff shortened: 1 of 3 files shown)


Audiobook Player/PlayerManager.swift
```diff
@@ -140,7 +144,9 @@ class PlayerManager: NSObject {
                 //set speed for player
                 audioplayer.rate = self.getSpeed()
 
-                NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.bookReady, object: nil)
+                DispatchQueue.main.async {
```
GianniCarlo: This dispatch shouldn't be necessary, it's already inside a dispatch main async 🤔

Audiobook Player/PlayerManager.swift
```diff
@@ -372,7 +378,9 @@ extension PlayerManager: AVAudioPlayerDelegate {
         if storedPercentage != percentageString {
             defaults.set(percentageString, forKey: self.identifier+"_percentage")
 
-            NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.updatePercentage, object: nil, userInfo: userInfo)
+            DispatchQueue.main.async {
```
GianniCarlo: `updateTimer` should already be on the main thread due to the timer being added to the `RunLoop.main`. I did some testing inserting `if Thread.isMainThread { print("on main thread") }` and it prints on every tick. Let me know in which specific case these notifications have been going on a background thread

Audiobook Player/PlayerManager.swift
```diff
@@ -386,11 +394,15 @@ extension PlayerManager: AVAudioPlayerDelegate {
 
             // Once book a book is finished, ask for a review
             defaults.set(true, forKey: "ask_review")
-            NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.bookEnd, object: nil)
+            DispatchQueue.main.async {
```
GianniCarlo: same comment as the last one

Audiobook Player/PlayerManager.swift
```diff
@@ -386,11 +394,15 @@ extension PlayerManager: AVAudioPlayerDelegate {
...
+                NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.bookEnd, object: nil)
+            }
         }
 
         //notify
-        NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.updateTimer, object: nil, userInfo: userInfo)
+        DispatchQueue.main.async {
```
GianniCarlo: same comment as the last one

Audiobook Player/PlayerManager.swift
```diff
@@ -402,11 +414,12 @@ extension PlayerManager: AVAudioPlayerDelegate {
             if Int(audioplayer.currentTime) >= chapter.start {
                 self.currentChapter = chapter
                 let chapterString = "Chapter \(chapter.index) of \(self.chapterArray.count)"
-                //notify
                 let userInfo = ["chapterString": chapterString, "fileURL": self.currentBooks.first!.fileURL] as [String: Any]
 
                 //notify
-                NotificationCenter.default.post(name: Notification.Name.AudiobookPlayer.updateChapter, object: nil, userInfo: userInfo)
+                DispatchQueue.main.async {
```
GianniCarlo: `updateCurrentChapter` is called from inside `updateTimer` so it shouldn't have a problem either

GianniCarlo:
@e7mac I left some comments regarding the dispatch.main.async, I've tested some of those notifications and they should be already on the main thread, let me know in which cases that isn't the case

e7mac:
@GianniCarlo those DispatchSync's were unnecessary. I put them as more of a style thing accompanying every Notification call rather than having to follow the code flow. However, updated the commit to mostly fix the bug only.

GianniCarlo:
@e7mac thanks again for your contribution! if you can, take a look at `next/major` branch, that'd be the base for the next release 3.0 👌

PR #75, opened 2018-05-02
Feature: New Player UI

pichfl:
1. New color system
2. New interactions
3. New icons
4. New toolbar

Re #44


Audiobook Player/Player/Containers/PlayerControlsViewController.swift
```diff
@@ -8,69 +8,225 @@
...
 import UIKit
 
-class PlayerControlsViewController: PlayerContainerViewController {
+class PlayerControlsViewController: PlayerContainerViewController, UIGestureRecognizerDelegate {
+    @IBOutlet private weak var artworkView: UIView!
     @IBOutlet private weak var artwork: UIImageView!
     @IBOutlet private weak var playPauseButton: UIButton!
-    @IBOutlet private weak var rewindButton: UIButton!
-    @IBOutlet private weak var forwardButton: UIButton!
+    @IBOutlet private weak var rewindIcon: PlayerForwardIconView!
```
GianniCarlo: I think `rewind`and `forward` references are swapped in the storyboard, right now the player controls are showing like this `-30` `play/pause` `+30`

GianniCarlo: updated comment

Audiobook Player/Player/Containers/PlayerControlsViewController.swift
```diff
@@ -8,69 +8,225 @@
...
-    @objc func onBookPause() {
+    @objc private func onBookPause() {
         self.isPlaying = false
     }
+
+    // MARK: Gesture recognizers
+
+    private func setupGestures() {
+        self.pan = UIPanGestureRecognizer(target: self, action: #selector(handlePan))
+        self.pan!.delegate = self
```
GianniCarlo: is `!` necessary?

Audiobook Player/Player/Containers/PlayerControlsViewController.swift
pichfl: 🤦‍♂️ they are

Audiobook Player/Player/Containers/PlayerControlsViewController.swift
pichfl: Nope, I think this is left from earlier code.

GianniCarlo:
I've got a suggestion for an improvement, when you drag down the player as if to dismiss it, currently you see a black view on the back, it'd be nice to be able to see the navigation bar of the previous VC. I think at some point I did something like this before in this app 🤔. Let me know what you think

Audiobook Player/Player/Containers/PlayerProgressViewController.swift
```diff
@@ -43,19 +43,31 @@ class PlayerProgressViewController: PlayerContainerViewController {
         return duration
     }
 
-    var tintColor: UIColor = .blue {
+    var colors: [UIColor]? {
         didSet {
-            self.progressSlider.tintColor = self.tintColor
+            guard let color = self.colors?[2] else {
```
GianniCarlo: magic number `[2]`

GianniCarlo: It'd be nice if we could structure somehow the array of colors returned by `CCColorCube`, so that it's clear that the first color is used by the background only and so on with the other colors

GianniCarlo: This comment about magic numbers applies to the other uses of the array colors too like in the `PlayerMetaViewController`, it seems a bit random for that class to know that author and chapter colors are stored in the index 2 and 3 out of that array of who knows how many more elements it may have

Audiobook Player/Player/Containers/PlayerJumpIconView.swift
```diff
@@ -0,0 +1,85 @@
...
+//  PlayerJumpIconView.swift
+//  Audiobook Player
+//
+//  Created by Florian Pichler on 22.04.18.
+//  Copyright © 2018 Tortuga Power. All rights reserved.
+//
+
+import UIKit
+
+enum PlayerJumpIconType {
```
GianniCarlo: Is this being used somewhere?

Audiobook Player/Player/Containers/PlayerControlsViewController.swift
```diff
@@ -8,69 +8,225 @@
...
+    private let jumpIconAlpha: CGFloat = 0.15
+    private var triggeredPanAction: Bool = false
+
+    private var isPlaying: Bool = false {
+        didSet {
+            self.playPauseButton.alpha = 1.0
+            self.playPauseButton.setImage(self.isPlaying ? self.pauseImage : self.playImage, for: UIControlState())
+
+            self.view.layoutIfNeeded()
+            self.artworkHeight.constant = self.isPlaying ? self.originalHeight : self.originalHeight * 255/325
```
GianniCarlo: what is `255` or `325` about?

Audiobook Player/Player/Containers/PlayerProgressViewController.swift
pichfl: To be honest, all colors after index 0 are non deterministic. 
But we can put them in a struct and apply arbitrary names which align with how we use them. 

Making the values a bit less magical is definitely good 🙂

Audiobook Player/Player/PlayerViewController.swift
```diff
@@ -50,6 +56,19 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
             }
     }
 
+    // Prevents dragging the view down from changing the safeAreaInsets.top
+    // Note: I'm pretty sure there is a better solution for this that I haven't found yet - @pichfl
+    override func viewSafeAreaInsetsDidChange() {
+        if #available(iOS 11, *) {
```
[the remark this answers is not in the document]
pichfl: The point is that there is no device before iOS 11 (which was the first OS released for the iPhone X) which actually needs this, which is why I went with the flag as the safe area insets are the preferred way by Apple to accommodate for the notch on the X.

pichfl:
Regarding the background: I opted for the default segue when I removed the deck transition lib. Adding a custom segue which renders the background sounds nice.

Audiobook Player/Player/PlayerViewController.swift
```diff
@@ -79,28 +99,68 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
         self.progressViewController?.book = currentBook
         self.progressViewController?.currentTime = UserDefaults.standard.double(forKey: currentBook.identifier)
 
-        self.speedButton.title = "\(String(PlayerManager.sharedInstance.speed))x"
+        self.speedButton.title = self.formatSpeed(PlayerManager.sharedInstance.speed)
 
-        // Colors
-        guard var artworkColors = NSArray(ofColorsFrom: currentBook.artwork, withFlatScheme: false) as? [UIColor] else {
+        guard let artwork: UIImage = currentBook.artwork else {
```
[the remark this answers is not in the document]
pichfl: I thought it would be wise to have the type for any declared variable, but I agree it should be obvious enough

Audiobook Player/Player/Containers/PlayerJumpIconView.swift
pichfl: Was replaced by the two subclasses below, will remove.

Audiobook Player/Player/Containers/PlayerControlsViewController.swift
pichfl: Magic value extracted from the design to determine how big the artwork should be for the scaled state. I should add a comment to explain or move the values to variables explaining what they represent (325 being the width of the artwork in the design in its default state and 255 being the scaled artwork)

Audiobook Player/Player/PlayerViewController.swift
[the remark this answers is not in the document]
GianniCarlo: True, I was basing my assumption on seeing this behavior in an iPhone 6s with iOS 11. I just tested it out in a simulator running iOS 10 and everything is good 👌

GianniCarlo:
Awesome! I think we're ready to merge

Issue #78, opened 2018-05-05
Next/Major: Play / Pause icon does not show up when returning to the Player screen


pichfl:
> When you open up the player, pause it, dismiss it, and open the player again on the same book. The book won’t start playing (as expected) but, the artwork doesn’t have the play button overlay

I think I'm not resetting the icons alpha when the view is presented.

PR #79, opened 2018-05-05
Feature: New default artwork

pichfl:
- Added new artwork
- Added new play / pause icon matching the new design
- Fixing #78 along the way.


GianniCarlo:
Awesome! great work @pichfl !

Issue #68, opened 2018-04-14
Move registering for remote events out of LibraryViewController


pichfl:
We should move this to a separate service class which in turn can be called from AppDelegate, which seems like a more appropriate place to me. But I can be wrong, please advise. 

Goal should be to make the purpose of the LibraryViewController more distinct to make it easier for the new features to be implemented.

GianniCarlo:
Agreed, AppDelegate sounds good for now. Besides moving the remote events, we could also move these events:
- `AVAudioSessionInterruption`
- `AVAudioSessionRouteChange`

Issue #52, opened 2018-03-29
New Miniplayer


pichfl:
Extracted from #46 as it makes sense to have this as a separate ticket.

- Now playing should use the same label behavior as the player screen (unlike the screenshot, can update if necessary): If a title is too long, it should fade out and scroll in intervals
- Now playing should behave like the download prompts in the iOS 11 App Store app: Translucent with blur and inverted for the play/pause mode. Make the tap targets as big as possible, whole square on the right side should react for play/pause and everything else should bring up the player.
- Optional animation: Now playing snaps into place from the bottom of the screen
- Assets will be provided by @pichfl on request

### Additional tasks

- [x] Animate and enable transition to player using a pan gesture
- [x] Color background and labels using the extracted colors
- [x] ~Blur effect~

PR #85, opened 2018-05-10
Feature: New miniplayer

pichfl:
Re #52 

The colors and the background are postponed until we finish the new data structure which will store the extracted colors.

This also improves the gestures for interacting with the player by adjusting their degrees of freedom and required distance. 

Re #68 - As I was messing with the LibraryViewController already, I also removed the Remote events to the app delegate.


BookPlayer/Credits.html
```diff
@@ -45,7 +45,7 @@
             @bryanrezende for the Smart rewind<br />
             @e7mac for the speed controls and autoplay<br />
             @gpambrozio for the Volume Boost<br />
-            @pichfl for Design, App Icon and Refactoring<br />
+            @pichfl for Design, Artwork and Refactoring<br />
```
GianniCarlo: same clarification should be on the `README.md`

GianniCarlo:
I just got one comment, which is arguable 🤔. If I tap the play/pause button on the mini player, the playerVC is shown, I could make an argument that if I took the trouble to actually hit the button to pause it, instead of opening the player first and pause it, I wouldn't want the player to be shown

pichfl:
I thought I had that added to the gestures, seems I stashed it at some point and forgot to commit

GianniCarlo:
There's a problem, when I drag the mini player, once the player is presented, if I haven't lifted my finger, and if I continue to drag, the app will try to show the player over and over again. This won't have any visual effect, but it will generate warning logs:
```
BookPlayer[15756:2196696] Warning: Attempt to present <BookPlayer.PlayerViewController: 0x1043440f0>  on <BookPlayer.LibraryViewController: 0x1043239a0> which is already presenting (null)
```
I found this while I was trying out a haptic feedback when the show player event was triggered from dragging the mini player
```
if #available(iOS 10.0, *) {
                    let generator = UINotificationFeedbackGenerator()
                    generator.notificationOccurred(.success)
                }
```
I think the feedback adds a nice touch to it, test it out and let me know what you think

pichfl:
I'll implement a similar reset mechanism to what I use for artwork. The vibration is a nice touch, I'll add it as well.

PR #88, opened 2018-05-10
Feature: Support section

pichfl:
**Sorry:** Please merge #85 before this one, I forgot to start from next/major when working on this.

Includes DeviceKit to get a human readable hardware identifier from the hardware version. Adds email and GitHub as possible support options, making it as easy as possible to send actionable problems. I also render the current version of the app in the section footer.

The email snippet contains the version of BookPlayer, the device and the OS version as well as a writing promt. We could consider extending this to include the usual bug report taglines in the future but until we get the first emails I'd stay with this simple version.

If no email account is available the info and email can be copied to the clipboard from the alert dialog.

This also includes a new run script which increases the build number automatically.


BookPlayer/Settings/SettingsViewController.swift
```diff
@@ -7,58 +7,119 @@
...
+            mail.setMessageBody("<p>Hello BookPlayer Crew,<br>I have an issue concerning BookPlayer \(appVersion) on my \(device) running \(systemVersion)</p><p>When I try to…</p>", isHTML: true)
+
+            self.present(mail, animated: true)
+        } else {
+            let debugInfo = "BookPlayer \(appVersion)\n\(device) with \(systemVersion)"
+
+            let alert = UIAlertController(title: "Unable to compose email", message: "You need to set up an email account in your device settings to use this. \n\nPlease mail us at \(self.supportEmail)\n\n\(debugInfo)", preferredStyle: .alert)
+
+            alert.addAction(UIAlertAction(title: "Copy information to clipboard", style: .default, handler: { _ in
+                UIPasteboard.general.string = "\(self.supportEmail)\(debugInfo)"
```
GianniCarlo: This needs a `\n` between the `supportEmail` and `debugInfo` otherwise when pasted, the string would be `support@bookplayer.appBookPlayer 3.0.0...`

Issue #71, opened 2018-04-29
Limit progress slider to current chapter


pichfl:
To make the control more useful we should consider limiting the progress slider to the current chapter. This makes more precise navigation easier. 

If a file has no chapters nothing would change. 

Its quite common in other apps (iBooks, Audible and a few other players I tested)

GianniCarlo:
We’ll have to adjust the labels too as audible does. I know most will welcome this change 👍

PR #89, opened 2018-05-12
Feature: Limit progress slider to current chapter for books with chapters

pichfl:
Re #71

### DIFF
--- a/BookPlayer/AppDelegate.swift
+++ b/BookPlayer/AppDelegate.swift
@@ -13,7 +13,7 @@ import Crashlytics
 import MediaPlayer
 
 @UIApplicationMain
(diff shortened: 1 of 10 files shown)


GianniCarlo:
Just one comment, if I drag the slider up to the end and hold it, the player will start skipping in a rapid way the chapters. I think we should imitate Audible's interaction, that you may drag it to the end of the chapter, but only when the finger is raised, the following chapter is started. Thoughts?

pichfl:
That will be an interesting thing to solve, but I agree.

pichfl:
I'm still missing one edge case, please wait with the merge

GianniCarlo:
awesome!

Issue #94, opened 2018-05-16
Status bar color not always in sync with screen color


pichfl:
Switching from Player -> Chapters -> Player shows the wrong status bar color if the player has a dark background.

pichfl:
Reopend as the previous fix doesn't work with the new root view controller.

pichfl:
New try in #102

PR #95, opened 2018-05-16
Bugfix: Allow player modal to adjust statusbar style

pichfl:
Using `self.modalPresentationCapturesStatusBarAppearance` was missing from the modal VC which is why setting `preferredStatusBarStyle` didn't work as expected.

Re #94


GianniCarlo:
nice!

PR #126, opened 2018-06-01
Cleanup: Extract Player into separate storyboard 

pichfl:
This should make future work on the player a bit easier to merge.

Please merge after #124 (I'll rebase as necessary), I started both at the same time and separating them would take more time than a quick rebase after #124 is merged.


pichfl:
@GianniCarlo Rebased 🙂

PR #128, opened 2018-06-03
Bugfix: Reset chapter cells on reuse

pichfl:
This also combines the duration formatters that were spread over the project and updates the layout for the chapter selector cells because a single line change would have been to easy.

Bugfix here:
https://github.com/pichfl/Audiobook-Player/blob/55b4131017fb8525065828ff4147d45bcfb348cd/BookPlayer/Player/ChaptersViewController.swift#L54


GianniCarlo:
nice!

Issue #129, opened 2018-06-03
Feature: Align Library layout with mockups


pichfl:
- [x] Playlists should have a disclosure indicator
- [x] Adjust label sizes 
- [x] Adjust alignment

Postponed until #131: Playlists should calculate and show an accumulated completion state.

PR #132, opened 2018-06-03
Feature: Match Library and playlists tableviews with design

pichfl:
Re #129


GianniCarlo:
There's a problem with the play/pause icon in the mini player
img_1021

pichfl:
Seems I failed to fix the constraints, on it.

Issue #73, opened 2018-04-30
New Empty State for Library and Playlists


pichfl:
Both the new Library screen and a freshly created Playlist need a good looking and helpful placeholder that shows up when no book was added yet.

This replaces #72 and extends on #46 

I'll add mockups as soon as possible.

pichfl:
I added artwork and views for both, only the "Add" cell is missing. Is there a way to reuse the UITableCell or is it not worth the effort?

GianniCarlo:
Since it's registered in the BaseListVC it shouldn't be a major problem. Just to be clear, in the library, when it's empty, the "Add" cell will be visible alongside the default artwork on the back right?

pichfl:
Here are two mockups for the screens. The "Add" button is centered in both. I'm not yet sure about the text for the Playlist button, but I think file is more appropriate in that context. 

### Empty Library

![library](https://user-images.githubusercontent.com/194641/40303276-7109f6aa-5cf2-11e8-82a6-41c9ed46fede.png)

### Empty Playlist 
![library empty](https://user-images.githubusercontent.com/194641/40303266-68011d2c-5cf2-11e8-9de3-4bccaccd7c86.png)

PR #133, opened 2018-06-03
Feature: Improve empty states for library and playlists

pichfl:
Re #73


GianniCarlo:
Possibly too much text, we could use this space just as a hint for the dragging functionality. I think we could do without the explicit part about importing files and maybe let the user find out the part where importing an existing file just moves the book around 🤔
img_bb12c7ce1d6b-1

pichfl:
Agreed. I'll update the PR with a shorter text.

GianniCarlo:
Do we need the spacer views? wouldn't it be cleaner to have the images center themselves vertically in relation to their superview. Maybe put an offset of `-41` to simulate the current position that their in. With that in place, the spacer views in both the player and playlist VCs could be deleted with no conflict

pichfl:
The spacers are used because I want to keep the views in balance when the length of the help text changes which depends on the system font size and/or localization. It's not necessary for the library but I thought keeping both views similar didn't hurt. We could consider extracting the empty state into a xib later on.

PR #146, opened 2018-06-09
Last book played

GianniCarlo:
- On launch, show player with last book played if it exists in UserDefaults
- Remove last book played from UserDefaults when the book is finished


GianniCarlo:
thank God for tests

PR #157, opened 2018-06-18
Internal: Custom app icons for debug and beta builds

pichfl:


BookPlayer.xcodeproj/project.pbxproj
```diff
@@ -793,7 +796,7 @@
 					"$(PROJECT_DIR)",
 				);
 				INFOPLIST_FILE = BookPlayer/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.3;
```
GianniCarlo: you need to set the same deployment target for the target `BookPlayerTests` then travis won't complain

pichfl: Done.

Issue #104, opened 2018-05-18
Books sometimes start erroneously from the beginning


pichfl:
It may be an development only issue but since we moved to CoreData books sometimes start from the beginning instead of the last saved position. Needs investigation.

GianniCarlo:
The other day I jumped back from the lock screen and it took me to the end (or was it the beginning? 🤔) it definitely needs investigation

pichfl:
I wonder if it is related to #106

GianniCarlo:
looks related to #108

pichfl:
I wasn't able to reproduce this for a while now, maybe it got fixed along the way?

pichfl:
Nope, was able to reproduce today. Started an audiobook using the play action on the playlist (which looks for the first unfinished book) and it started the first book in a list of three from start instead of from about the middle. 

There must be an error in how we save, retrieve or calculate the last played position. No idea why

PR #161, opened 2018-06-18
Bugfix: Call update before pausing to ensure saving

pichfl:
Stopping the player before the update can unravel the player to a point where it makes the update method bail early on instead of saving the current position.

Re #104

### DIFF
--- a/BookPlayer/Player/PlayerManager.swift
+++ b/BookPlayer/Player/PlayerManager.swift
@@ -10,6 +10,8 @@ import Foundation
 import AVFoundation
 import MediaPlayer
 
+// swiftlint:disable file_length
+
 class PlayerManager: NSObject {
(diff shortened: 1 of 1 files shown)


BookPlayer/Player/PlayerManager.swift
```diff
@@ -10,6 +10,8 @@ import Foundation
 import AVFoundation
 import MediaPlayer
 
+// swiftlint:disable file_length
```
GianniCarlo: travis is failing due to swiftlint, `PlayerManager` isn't over 400 lines, so we can delete this comment safely

pichfl: Artifact from extracting this PR from my other work, should be fixed now.

Issue #155, opened 2018-06-18
Player does not resume playback in all cases


pichfl:
Check for proper handling of the `shouldResume` in notification. Likely related to #151 

- Setting a Timer in the Clock.app and having it gone off will pause but not unpause the player
- Same with playing and stopping a video in youtube: Player pauses but does not unpause
- Bringing up Siri has the same effect

I think the latter happens in other audio apps as well, but the former should not as the timer is merely an interruption and not a change in medium

GianniCarlo:
In which specific cases has this happened? not sure what to test yet

GianniCarlo:
#174 resolves all those points, except the youtube one, since that interruption isn't an 'unexpected' one, the app doesn't get the callback when the youtube video is paused

pichfl:
Can confirm, #174 fixes the interruptions for Apple Maps as well :)

PR #174, opened 2018-06-24
Fix handling resuming audio after interruptions

### LINKED ISSUE #151: Audio not pausing when Apple Maps giving directions
Apple Maps has a feature that allows apps that designate themselves as “spoken word audio” to be automatically paused when Maps is speaking directions in navigation. BookPlayer is not pausing when directions are being said in Maps, instead continuing to play in the background.

GianniCarlo:
This should fix #151 (tested with Google Maps) and #155

### DIFF
--- a/BookPlayer/Player/PlayerViewController.swift
+++ b/BookPlayer/Player/PlayerViewController.swift
@@ -185,7 +185,7 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
             },
             onEnd: { (_ cancelled: Bool) -> Void in
                 if !cancelled {
-                    PlayerManager.shared.stop()
+                    PlayerManager.shared.pause()
                 }
 
                 self.sleepLabel.title = ""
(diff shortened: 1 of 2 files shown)


BookPlayer/Player/PlayerViewController.swift
```diff
@@ -185,7 +185,7 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
             },
             onEnd: { (_ cancelled: Bool) -> Void in
                 if !cancelled {
-                    PlayerManager.shared.stop()
+                    PlayerManager.shared.pause()
```
GianniCarlo: thoughts?

pichfl: Totally for it. Given our current interaction model, `stop()` is used only when a book is about to be deleted. Or when it's paused and the mini player is dismissed. Pausing makes much more sense for the timer :)

PR #175, opened 2018-06-24
Reset smart rewind on scrubbing

### LINKED ISSUE #168: Smart rewind is applied even when the play position was changed before beginning playback
To reproduce this pause the book, wait some time, scrub using the progress slider and start playback again.

Most likely fix: Resetting the smart rewind when the progress was changed by `PlayerManager.jumpBy()`

GianniCarlo:
- Resolve #168
- Change rewind from `60` to `30`, it's happened to me that I have actually been constantly going backwards in a book thanks to the smart rewind 🤔

### DIFF
--- a/BookPlayer/Player/PlayerManager.swift
+++ b/BookPlayer/Player/PlayerManager.swift
@@ -25,7 +25,7 @@ class PlayerManager: NSObject {
 
     // 599 = 10 mins
     private let smartRewindThreshold = 599.0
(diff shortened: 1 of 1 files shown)


pichfl:
I agree with reducing the maximum amount, the 60s limit was a bit high when in use. I have created #177 to introduce a different calculation of the amount to rewind by, which should feel a little less intrusive.

Issue #112, opened 2018-05-22
Dismiss mini player


pichfl:
1. It would be nice to be able to dismiss the mini player by swiping horizontally - it is not necessary, but it would be a nice touch.
2. the mini player must be dismissed when the currently playing book is deleted
3. To dismiss the player, the player must be paused
4. Dismissing the mini player will stop the audio player session

PR #179, opened 2018-06-24
 Feature: Allow dismissal of paused Mini Player

pichfl:
- Dismissing the Mini Player by swiping down stops the book
- Stopping a book now removes the playing state from its cell
- A playing book can not be dismissed

Re #112

### DIFF
--- a/BookPlayer/Library/BaseListViewController.swift
+++ b/BookPlayer/Library/BaseListViewController.swift
@@ -59,6 +59,7 @@ class BaseListViewController: UIViewController {
         NotificationCenter.default.addObserver(self, selector: #selector(self.presentMiniPlayer), name: Notification.Name.AudiobookPlayer.playerDismissed, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(self.onBookPlay), name: Notification.Name.AudiobookPlayer.bookPlayed, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(self.onBookPause), name: Notification.Name.AudiobookPlayer.bookPaused, object: nil)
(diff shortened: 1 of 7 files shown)


BookPlayer/Library/RootViewController.swift
```diff
@@ -91,28 +105,102 @@ class RootViewController: UIViewController {
...
         guard let userInfo = notification.userInfo,
             let book = userInfo["book"] as? Book else {
                 return
         }
 
-        setupFooter(book: book)
+        setupMiniPlayer(book: book)
+    }
+
+    @objc private func bookDeleted(_ notification: Notification) {
```
GianniCarlo: it seems like the notification `bookDeleted` isn't being actually used

BookPlayer/Library/BaseListViewController.swift
```diff
@@ -81,6 +82,19 @@ class BaseListViewController: UIViewController {
         bookCell.playbackState = .paused
     }
 
+    @objc func onBookStop(_ notification: Notification) {
+        guard
+            let userInfo = notification.userInfo,
+            let book = userInfo["book"] as? Book,
+            let index = self.library.itemIndex(with: book.fileURL),
```
GianniCarlo: This needs to be overridden in the `PlaylistViewController` to get the index inside the playlist

BookPlayer/Library/RootViewController.swift
pichfl: 😯 oh right, I forgot the second part of the ticket. Deleting a book should stop the book as well.

BookPlayer/Library/BaseListViewController.swift
pichfl: Done.

PR #181, opened 2018-06-24
Release 3.0.0

GianniCarlo:
This wouldn't have been possible without you @pichfl , behold all those commits since April 💪
This marks the end of development for version 3.0.0 🎉.


pichfl:
I'm not quite sure on the best way to review this. I've put this branch on my test devices and will be running it for some time.

GianniCarlo:
I think the testing on each previous PR should be enough, we'll setup beta testing with testflight prior to the appstore release

Issue #182, opened 2018-06-25
Removing a book from the Library does not dismiss the Mini Player


pichfl:
I thought I had this one fixed already, seems I forgot to set some state.

PR #184, opened 2018-06-25
Bugfix: Deleting a book should dismiss Mini Player

pichfl:
More precisely: Calling stop() should send a notification in any case, which in turn will dismiss the Mini Player if the stop() happend for the current book

Re #182

### DIFF
--- a/BookPlayer/Player/PlayerManager.swift
+++ b/BookPlayer/Player/PlayerManager.swift
@@ -373,18 +373,29 @@ class PlayerManager: NSObject {
     }
 
     func stop() {
+        self.audioPlayer?.stop()
+
         guard let book = self.currentBook else {
+            DispatchQueue.main.async {
+                NotificationCenter.default.post(
+                    name: Notification.Name.AudiobookPlayer.bookStopped,
+                    object: nil
+                )
+            }
+
             return
         }
 
-        self.audioPlayer?.stop()
+        let userInfo = ["book": book]
 
         self.currentBooks = []
 
-        let userInfo = ["book": book]
-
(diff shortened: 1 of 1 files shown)


BookPlayer/Player/PlayerManager.swift
```diff
@@ -373,18 +373,12 @@ class PlayerManager: NSObject {
...
     func stop() {
-        guard let book = self.currentBook else {
-            return
-        }
-
         self.audioPlayer?.stop()
 
         self.currentBooks = []
 
-        let userInfo = ["book": book]
```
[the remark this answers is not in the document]
pichfl: 🤦‍♂️

BookPlayer/Player/PlayerManager.swift
```diff
@@ -373,18 +373,29 @@ class PlayerManager: NSObject {
     }
 
     func stop() {
+        self.audioPlayer?.stop()
+
         guard let book = self.currentBook else {
```
GianniCarlo: if the notification will occur no matter what, let's change this function to the following then:
```
self.audioPlayer?.stop()
var userInfo: [AnyHashable: Any]?

if let book = self.currentBook {
  userInfo = ["book":book]
}

self.currentBooks = []

DispatchQueue.main.async {
            NotificationCenter.default.post(
                name: Notification.Name.AudiobookPlayer.bookStopped,
                object: nil,
                userInfo: userInfo
            )
}
```

pichfl: That's what I was looking for. Much more reasonable.

Issue #188, opened 2018-06-26
UI constraints missing for iPhone Plus


pichfl:
Some elements on the Player Screen have the wrong width

pichfl:
Settings as well: ![image 2018-06-26 22 11 59](https://user-images.githubusercontent.com/194641/41936602-f782706a-798d-11e8-964d-32b6298fef12.jpg)

PR #189, opened 2018-06-26
Bugfix: Adapt layout for all iPhone screen sizes

pichfl:
Re #188

### DIFF
--- a/BookPlayer/Library/Views/BookCellView.swift
+++ b/BookPlayer/Library/Views/BookCellView.swift
@@ -47,6 +47,9 @@ class BookCellView: UITableViewCell {
             } else if ratio < 1 {
                 self.artworkHeight.constant = 50.0
                 self.artworkWidth.constant = 50.0 * ratio
+            } else {
+                self.artworkHeight.constant = 50.0
+                self.artworkWidth.constant = 50.0
             }
         }
     }
--- a/BookPlayer/Library/Views/BookCellView.swift
+++ b/BookPlayer/Library/Views/BookCellView.swift
@@ -41,16 +41,8 @@ class BookCellView: UITableViewCell {
 
             let ratio = self.artworkView.imageRatio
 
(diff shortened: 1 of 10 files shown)


BookPlayer/Library/Views/BookCellView.swift
```diff
@@ -47,6 +47,9 @@ class BookCellView: UITableViewCell {
             } else if ratio < 1 {
                 self.artworkHeight.constant = 50.0
                 self.artworkWidth.constant = 50.0 * ratio
+            } else {
```
GianniCarlo: I think we can replace this `if ... else if ... else` for these two lines:
```
self.artworkHeight.constant = ratio > 1 ? 50.0 / ratio ? 50
self.artworkWidth.constant = ratio < 1 ? 50.0 * ratio ? 50
```

BookPlayer/Player/Views/ArtworkControl.swift
```diff
@@ -54,10 +60,23 @@ class ArtworkControl: UIView, UIGestureRecognizerDelegate {
...
         }
+
         set {
             self.artworkImage.image = newValue
 
-            self.artworkImage.resizeToFit()
+            let ratio = self.artworkImage.imageRatio
+            let base = min(self.artworkContainer.bounds.width, self.artworkContainer.bounds.height)
+
+            if ratio > 1 {
```
GianniCarlo: Same comment as before, I think we can replace this `if ... else if ... else` for these two lines:
```
self.artworkWidth.constant = ratio > 1 ? base / ratio ? base
self.artworkHeight.constant = ratio < 1 ? base * ratio ? base
```

BookPlayer/Library/Views/BookCellView.swift
pichfl: Agreed.

PR #195, opened 2018-07-02
Fix crash on migration

GianniCarlo:
- Adding a queue fixes the crash of multiple insertions at the same time
- Fix legacy identifier for migration old book progress


GianniCarlo:
This resolves #194

Issue #196, opened 2018-07-02
3.0.0 beta: Add drag hint on player


GianniCarlo:
The skip back/forward  rework isn't apparent, we need to add a hint to let the user know that it's no longer a button

pichfl:
I had scheduled this for 3.1 in #158  but it seems logical to move this to 3.0. I adjusted the original ticket as necessary.

Todo: 
- Hint forward/rewind  if a user has never used it by nudging the artwork to the right
- Resize Play / Pause button action with the cover artwork, limiting the interactive area to the cover.
- Register taps on the jump icons to nudge the artwork to the side without actually navigating

PR #197, opened 2018-07-02
Feature: Hint cover drag for first time users

pichfl:
This enables a drag hint that nudges the artwork every time the player is presented until the user uses the drag gesture to rewind or forward.

The same nudge is used when the icons behind the artwork are tapped in an attempt to jump that way.

This should provide enough feedback and encouragement to users that return from an earlier version and also convey a little achievement when they interact with the artwork the first time.

Re #196

### DIFF
--- a/BookPlayer/Player/Containers/PlayerControlsViewController.swift
+++ b/BookPlayer/Player/Containers/PlayerControlsViewController.swift
@@ -62,10 +62,18 @@ class PlayerControlsViewController: PlayerContainerViewController, UIGestureReco
         return duration
     }
 
+    private var artworkJumpControlsUsed: Bool = false {
+        didSet {
+            UserDefaults.standard.set(self.artworkJumpControlsUsed, forKey: UserDefaultsConstants.artworkJumpControlsUsed)
+        }
+    }
+
     // MARK: - Lifecycle
 
     override func viewDidLoad() {
         super.viewDidLoad()
+
+        self.artworkJumpControlsUsed = UserDefaults.standard.bool(forKey: UserDefaultsConstants.artworkJumpControlsUsed)
 
         self.artworkControl.isPlaying = PlayerManager.shared.isPlaying
         self.artworkControl.onPlayPause = { control in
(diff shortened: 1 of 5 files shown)


pichfl:
I had to get a bit create to match the play/pause button to the artwork size in paused mode as simply scaling the button would have changed the scale of the play button which I didn't intend to. 

I'm setting a new key to user defaults and since I suspect user defaults to be a bit on the slow side, I only read and set them only indirectly. That may not be necessary, but I tried to match what we have elsewhere in the app.

BookPlayer/Player/Containers/PlayerControlsViewController.swift
```diff
@@ -86,8 +102,12 @@ class PlayerControlsViewController: PlayerContainerViewController, UIGestureReco
         NotificationCenter.default.addObserver(self, selector: #selector(self.onPlayback), name: Notification.Name.AudiobookPlayer.bookPlaying, object: nil)
     }
 
-    override func didReceiveMemoryWarning() {
-        super.didReceiveMemoryWarning()
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        if !UserDefaults.standard.bool(forKey: UserDefaultsConstants.artworkJumpControlsUsed) {
```
GianniCarlo: why not use `self.artworkJumpControlsUsed`? it's already intialized in `viewDidLoad`

BookPlayer/Player/Views/ArtworkControl.swift
```diff
@@ -145,17 +151,25 @@ class ArtworkControl: UIView, UIGestureRecognizerDelegate {
...
         self.pan.cancelsTouchesInView = true
 
         self.addGestureRecognizer(self.pan!)
-    }
 
-    override func layoutIfNeeded() {
+        self.tap = UITapGestureRecognizer(target: self, action: #selector(tapAction))
+        self.tap.delegate = self
+        self.tap.numberOfTapsRequired = 1
+        self.tap.numberOfTouchesRequired = 1
```
GianniCarlo: default value is already `1` https://developer.apple.com/documentation/uikit/uitapgesturerecognizer/1623580-numberoftouchesrequired

BookPlayer/Player/Containers/PlayerControlsViewController.swift
pichfl: Doh! Forgot to refactor that one

Issue #207, opened 2018-07-12
Deleting an empty playlist could forgo the alert


pichfl:
This is a small nitpick, but empty playlists ask just as full ones if books inside the playlist should be kept. I think we can remove them straight away.

PR #213, opened 2018-07-16
Feature: Delete empty playlists without confirmation

pichfl:
There is little effort involved in recreating an empty playlist so it makes sense to skip the confirmation dialog.

Re #207

### DIFF
--- a/BookPlayer/Library/LibraryViewController.swift
+++ b/BookPlayer/Library/LibraryViewController.swift
@@ -120,6 +120,20 @@ class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate
     }
 
     func handleDelete(playlist: Playlist, indexPath: IndexPath) {
+        if playlist.books?.count == 0 {
+            self.library.removeFromItems(playlist)
+
+            DataManager.saveContext()
+
+            self.tableView.beginUpdates()
+            self.tableView.deleteRows(at: [indexPath], with: .none)
+            self.tableView.endUpdates()
+
+            self.emptyLibraryPlaceholder.isHidden = !self.items.isEmpty
+
+            return
+        }
+
         let sheet = UIAlertController(
             title: "Delete \(playlist.title!)?",
             message: "Deleting only the playlist will move all its files back to the Library.",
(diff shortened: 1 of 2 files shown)


BookPlayer/Library/LibraryViewController.swift
```diff
@@ -120,6 +120,20 @@ class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate
     }
 
     func handleDelete(playlist: Playlist, indexPath: IndexPath) {
+        if playlist.books?.count == 0 {
```
GianniCarlo: change to `guard` syntax

BookPlayer/Library/LibraryViewController.swift
```diff
@@ -120,6 +120,20 @@ class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate
...
+            self.library.removeFromItems(playlist)
+
+            DataManager.saveContext()
```
GianniCarlo: I've seen that we use this sequence many times
`saveContext`
`beginUpdates`
`deleteRows`
`endUpdates`
We might want to put that in its own function that accepts an array of IndexPaths 🤔
Thoughts?

BookPlayer/Library/LibraryViewController.swift
```diff
@@ -219,7 +235,14 @@ extension LibraryViewController {
         let item = self.items[indexPath.row]
 
         // "…" on a button indicates a follow up dialog instead of an immmediate action in macOS and iOS
-        let deleteAction = UITableViewRowAction(style: .default, title: "Delete…") { (_, indexPath) in
+        var title = "Delete…"
+
+        // Remove the dots if trying to delete an empty playlist
+        if item is Playlist {
```
GianniCarlo: mmmm might as well unwrap it here and avoid 'forcing it' inside
```
if let playlist = item as? Playlist {
  title = (playlist.books?.count ?? 0) > 0 ? title: "Delete"
}
```
To be honest I'd like to replace `.count ?? 0) > 0 ` with a simple `!isEmpty` but that is just for an array not an ordered set 🤔

BookPlayer/Library/LibraryViewController.swift
pichfl: I guess you're right. I just didn't want to introduce too many new things in this, but it makes sense.

pichfl: After some thinking I'm no longer convinced we would win that much by creating an helper here. I think it will make that part more complex for brevities sake and if it comes to that I prefer the more easily readable solution.

PR #214, opened 2018-07-16
Bugfix: Use secondary views for empty state overlays

pichfl:
This makes the storyboard easier to read and properly shows the empty state when the last item was removed from a playlist.

### DIFF
--- a/BookPlayer/Library/LibraryViewController.swift
+++ b/BookPlayer/Library/LibraryViewController.swift
@@ -11,11 +11,7 @@ import MediaPlayer
 import MBProgressHUD
 import SwiftReorder
 
-// swiftlint:disable file_length
-
 class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate {
-    @IBOutlet private weak var emptyStatePlaceholder: UIView!
-
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -58,18 +54,6 @@ class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate
         //for iOS 8
         NotificationCenter.default.removeObserver(self)
         UIApplication.shared.endReceivingRemoteControlEvents()
-    }
-
-    fileprivate func toggleEmptyStateView() {
-        guard let placeholder = self.emptyStatePlaceholder else {
-            return
-        }
-
-        if self.items.isEmpty {
-            self.view.addSubview(placeholder)
-        } else {
-            placeholder.removeFromSuperview()
-        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


BookPlayer/Library/LibraryViewController.swift
```diff
@@ -58,15 +60,27 @@ class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate
         UIApplication.shared.endReceivingRemoteControlEvents()
     }
 
+    fileprivate func toggleEmptyStateView() {
```
GianniCarlo: This declaration can go inside `BaseListViewController` since the implementation for `LibraryViewController` isn't different than the on in `PlaylistViewController`

BookPlayer/Library/LibraryViewController.swift
```diff
@@ -58,15 +60,27 @@ class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate
...
+        guard let placeholder = self.emptyStatePlaceholder else {
+            return
+        }
+
+        if self.items.isEmpty {
```
GianniCarlo: Any particular reason as to why not just hide it instead of removing it from the superview?

BookPlayer/Library/LibraryViewController.swift
pichfl: I wish. I can't add this to the parents as the outlets referenced do not exist in the parent and I haven't figured out how to override outlets in subclasses

GianniCarlo: oh right 😅 my bad

BookPlayer/Library/LibraryViewController.swift
pichfl: The view isn't in the superview (see Storyboard), so I'd have to add it once in the viewDidLoad and I figured it would be cleaner that way, but if you think otherwise, we can change it back.

BookPlayer/Library/LibraryViewController.swift
GianniCarlo: although there might be something, let me test something out

GianniCarlo: alright:
1.- Move the `IBOutlet emptyStatePlaceholder` to `BaseListViewController` (delete the var from both Library and Playlist ViewControllers)
2.- From the storyboard reconnect both outlets from the two VCs into the one in the `BaseListViewController`
3.- Done

pichfl: Ohhh that works? 😅 Will fix asap.

BookPlayer/Library/LibraryViewController.swift
GianniCarlo: 🤔 but it *is* in the superview, for clarification's sake, I don't mean the `BaseListViewController`. What worries me about removing it from the superview instead of hiding it, is that it might turn into nil after removing it, since the superview is the one that holds the strong reference to that variable

pichfl: Hmm true.

Issue #218, opened 2018-07-18
Global autoplay


pichfl:
We have a few vocal requests to bring back global autoplay. While playlists fill that gap quite nicely, I don't think it will hurt to bring this back with these restrictions:

- Default state should be off, but the previously used flag can be reused
- Playlists should be queued as their files

Note to BookPlayer users: Please note, this is not a bug, but a new feature. The previous global playback was not prepared for the new Library format or Playlists.

PR #221, opened 2018-07-22
Feature: Autoplay Library

pichfl:
Re #218

### DIFF
--- a/BookPlayer/Library/BaseListViewController.swift
+++ b/BookPlayer/Library/BaseListViewController.swift
@@ -156,15 +156,16 @@ class BaseListViewController: UIViewController {
 
     func queueBooksForPlayback(_ startItem: LibraryItem, forceAutoplay: Bool = false) -> [Book] {
         var books = [Book]()
-        let isPlaylist = startItem is Playlist
         let shouldAutoplayLibrary = UserDefaults.standard.bool(forKey: UserDefaultsConstants.autoplayEnabled)
-        let shouldAutoplay = (isPlaylist || shouldAutoplayLibrary || forceAutoplay)
+        let shouldAutoplay = shouldAutoplayLibrary || forceAutoplay
 
-        guard let book = startItem as? Book else {
-            return books
+        if let book = startItem as? Book {
+            books.append(book)
         }
 
-        books.append(book)
+        if let playlist = startItem as? Playlist {
+            books.append(contentsOf: playlist.getRemainingBooks())
+        }
 
         guard
             shouldAutoplay,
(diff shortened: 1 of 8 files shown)


BookPlayer/Library/BaseListViewController.swift
```diff
@@ -152,23 +154,57 @@ class BaseListViewController: UIViewController {
...
+        guard
+            shouldAutoplay,
+            let remainingItems = self.items.split(whereSeparator: { $0 == startItem }).last
+            else {
+                return books
+        }
+
+        for item in remainingItems {
+            if let playlist = item as? Playlist, let playlistBooks = playlist.books?.array as? [Book] {
+                for book in playlistBooks {
```
GianniCarlo: instead of this `for` loop, you can do `books.append(contentsOf: playlistBooks)`

pichfl: Ever learning. Will update.

BookPlayer/Library/LibraryViewController.swift
```diff
@@ -302,22 +302,30 @@ extension LibraryViewController {
...
+        if let book = self.items[indexPath.row] as? Book {
+            let books = self.queueBooksForPlayback(book)
 
-                self.navigationController?.pushViewController(playlistVC, animated: true)
-            }
+            self.setupPlayer(books: books)
+        }
+    }
 
+    private func presentPlaylist(_ playlist: Playlist) {
```
GianniCarlo: testing webhooks in pr comments

GianniCarlo:
There's a inconsistency between the lock screen player and the current book playing 
img_1669
img_1670

GianniCarlo:
If the playthrough starts with a playlist (via artwork tap) the autoplay won't work

pichfl:
I think you may have discovered another bug with how changed information is sent to `MPNowPlayingInfoCenter`. I'll keep that separate from this PR but will look into it.

pichfl:
Before we merge this I we should discuss what autoplay actually does. Should it skip completed files? It does when starting a playlist.

GianniCarlo:
👌 sorry for the delay. For documentation/traceability sake: as we discussed (and already implemented in this PR) completed files will be skipped

pichfl:
This is also noted in the help text on the settings screen, see here:
https://github.com/TortugaPower/BookPlayer/blob/8742bd956dafed5a9510452c096e5ea5021ae9f3/BookPlayer/Base.lproj/Main.storyboard#L88

Issue #203, opened 2018-07-10
Support for controlling AirPlay playback


pichfl:
Many other players include an icon in their toolbar that allows one to control the currently selected AirPlay targets, I think we would be in good company to do so as well.

PR #226, opened 2018-08-12
Feature: Enable AirPlay route picker where available

pichfl:
This adds an additional button to the player toolbar which brings up the system dialog to select the AirPlay destination.

This also fixes a long stand bug that only disabled the chapter button instead of actually removing it (which came to be because I didn't understand `isEnabled` at the time).

This also introduces an animated change in the toolbar when the sleep timer is started or ended.

Re #203 

Signed-off-by: Florian Pichler <fp@ylk.gd>


BookPlayer/Player/PlayerViewController.swift
```diff
@@ -142,6 +139,48 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
         self.setNeedsStatusBarAppearanceUpdate()
     }
 
+    func updateToolbar(_ showTimerLabel: Bool = false, animated: Bool = false) {
+        guard var items = self.bottomToolbar.items else {
+            return
+        }
+
+        items = items.filter({ $0.tag > 0 })
```
GianniCarlo: Since we already have the `IBOutlets` for all the buttons (except for the 'more' button), we could have:
```
var items = [
  self.speedButton,
  self.sleepButton
]

if currentBook.hasChapters {
  items.append(self.chaptersButton)
}

if #available(iOS 11, *) {
 .
 .
  items.append(avRoutePickerBarButtonItem)
}
.
.
```
This way we avoid relying on tags and it looks less convoluted 🤔, since we won't be deconstructing and array, but building the array

GianniCarlo: with that approach we would have to add the 'more' button outlet though

GianniCarlo:
let me know what you think of the approach I mentioned in the comment, otherwise, this is working great 👌

BookPlayer/Player/PlayerViewController.swift
pichfl: Feels better that way. Less trying to bend things into the right shape, more building the right structure from ground up.

PR #233, opened 2018-09-05
Voice over support, resolves #205

### LINKED ISSUE #205: VoiceOver support
We should check VoiceOver and general accessibility guidelines for the application. @dunkelstern might be able to help with this.

ryantstone:
Adds voiceover support. Opening the PR for now, it could definitely use some cleanup but I wanted to get the PR open to discuss what changes need to be made.

References: #205

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: swift
-osx_image: xcode9.3
+osx_image: xcode10
 before_install:
   - brew update
   - brew outdated carthage || brew upgrade carthage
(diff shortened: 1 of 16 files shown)


GianniCarlo:
@ryantstone awesome! I'll check it out between today and tomorrow 👌

GianniCarlo:
I've been testing it out on my device, I got some points to discuss:
1.- think we should disable the artwork on the list as a accessibility element, so the user when going through the list will go through: item 1 -> item 2... instead of: item 1 -> artwork1 -> item 2 -> artwork2...
2.- we should enable the dismiss 'Z' gesture inside the playerView
Bonus
3.- make it so users can jump between only playlists or only books (it might be possible with the rotor thing)

GianniCarlo:
Let me know what you think @ryantstone

ryantstone:
Implemented book and playlist rotors, added Z gesture & removed the artwork when rolling through things.

GianniCarlo:
awesome! I'll test it out tomorrow

GianniCarlo:
I don't see the custom rotors :/ not sure why. Also, I think travis is failing due to `firstIndex(of:` requires Xcode 10, so we'll have to update `travis.yml` the line with `osx_image: xcode9.3` to `osx_image: xcode10`

ryantstone:
Ran into the no rotor issue too and thought it was just a problem with my machine. I would build and it wouldn't be there then I would rebuild and it would work. And I'll take care of the Travis file.

Issue #222, opened 2018-07-23
Sorting playlists alphanumerically


martinklepsch:
* **Feature Request**
* Files are ordered in a non-obvious way (order if import maybe?) 
* Most of the time filenames do contain some information about ordering, e.g. "Chapter 03".
* Manually ordering files is tedious and allowing users to sort them seems like a potential "quick win" (I'm not in a position to judge that however so YMMV)

---

This app is looking super promising! I've been waiting for a dead simple audiobook app that lets me just copy files via AirDrop or some similarly straightforward mechanism for ages. Would be happy to pay for the app or support a Patreon or similar. Great work! 🙌

pichfl:
First of all, I'm happy to hear you enjoy the app. We are considering taking donations, but we have to figure out the implications for this project.

Yes, you're correct. Files are sorted by import date by default. Adding other sorting options is totally possible and was even part of my initial mockups for version 3. 

To keep our backlog clean and focussed, I'm closing this ticket and adding it to our roadmap.

ryantstone:
@pichfl You think a `Sort` button on the playlist right corner bar that brings up a 2 option dialog at the bottom Title & Filename?

ryantstone:
I can build the sorting without a front end if you’d like to design something specific.

pichfl:
Nah, start with the button, we'll merge things down the line.

ryantstone:
I have a PR ready, which branch should I open it to merge into? `next/patch`?

GianniCarlo:
Awesome!, do it pointing to `develop` please

PR #237, opened 2018-09-13
Batch import, resolves #225

### LINKED ISSUE #225: Handle importing files as a batch operation
Each batch of files should:
- Prompt the user `(if files > 1)` to create a playlist which will contain such books
- Alphabetically sort the imported files

With this approach, it'll be easier to handle the feedback shown to the user with the new cell that @pichfl provided

GianniCarlo:
~~I still need to modify the existing tests~~, but this branch is already working and contains the new way to import as batch.

The process is as follows:
1- File is imported into the app (via airdrop/itunes/etc)
2.- A `newFileUrl` notification is generated and posted
3.- Either the library or the playlist VC (we'll call them the ListVC in this explanation) will handle this notification and tell the `DataManager` to process the url
4.- The `DataManager` holds a `ImportManager` instance, which is in charge to create a `ImportOperation` once a finite amount of time (1 sec) passes without a new `newFileUrl` notification ocurring (this is due to airdrop files being notified one at a time by the OS, instead of providing an array of urls).
5.- Once a `ImportOperation` is created, the  ListVC is notified and will set the completion block of the operation before notifying the `DataManager` to start the operation.
6.- Operation finishes, the ListVC will behave a bit different depending if it's the Library or the Playlist VC

Now there's an alert prompting you to create a new playlist with the batch of files recently imported


GianniCarlo:
There's a bug that I'll look into tomorrow, if you airdrop files while you're in the player view, neither the Library nor the Playlist VC will be subscribe to the event

GianniCarlo:
other things to do: 
- clean up files if they for whatever reason fail to either process or create a `Book` instance
- review loading view design

GianniCarlo:
I'll set it to 2 secs then 👌 I'm also changing to automatically inserting the books in the library and then presenting the playlist alert. This will help avoid getting the books stuck in limbo if for example another import operation is completed and the first alert is overridden by the new one. Although the proper way would be to tell the operation queue to hold on until the completed operation is 'processed' before starting the new one... Mmmmmmm might have to do it with operation dependencies 🤔

ryantstone:
I ran into an issue after the import. The order of the files wasn't as expected. The file structure was in correct order in the files app but the order of the playlist ordered by the mp3 title. Either here or in another issue we might want to add a button for reordering the playlist between file name and title.

pichfl:
@ryantstone I think #222 should be a start. Adding a "Sort" button with different options sounds more than reasonable.

ryantstone:
@pichfl I'll happily work on #222

GianniCarlo:
I'll check that out @ryantstone thanks for testing it. The batch-import should have the files sorted alphabetically, as a half-measure until something like #222 is implemented

GianniCarlo:
an update on what @ryantstone mentioned, when importing the files, the Book instance isn't created until it's verified that it's not already in the library, so we don't have access to the metadata yet to order it. For the time being the batch-import maintains the order based on the file title, if anyone wants to check out where the file where the [magic happens](https://github.com/TortugaPower/BookPlayer/pull/237/files#diff-6a2e7c37f2b69175eda47c3c16d56f5dR214)

PR #262, opened 2018-10-14
Sortable playlists

ryantstone:
This adds a sort order feature to the playlist view controller. Allows sorting by metadata title or original filename. It will display an error if previously imported files try to sort by filename since it's a new optional attribute on the book model. Let me know if any there you want any wording changes on the messages.

### DIFF
--- a/BookPlayer/Library/PlaylistViewController.swift
+++ b/BookPlayer/Library/PlaylistViewController.swift
@@ -10,42 +10,7 @@ import UIKit
 
 class PlaylistViewController: BaseListViewController {
     var playlist: Playlist!
-    @IBAction func didTapSort(_ sender: Any) {
-        let alert = UIAlertController(title: "Sort Files", message: "Sort Playlist files by", preferredStyle: .actionSheet)
-        alert.addAction(UIAlertAction(title: "Title", style: .default, handler: { (action) in
-            guard let books = self.playlist.books else { return }
-            do {
-                let sortedBooks = try BookSortService.init(books: books).perform(filter: .metadataTitle)
-                self.playlist.books = sortedBooks
-                DataManager.saveContext()
-                self.tableView.reloadData()
-            } catch {
-                print(error)
-            }
 
-        }))
-
-        alert.addAction(UIAlertAction(title: "File Name", style: .default, handler: { (action) in
-
-            guard let books = self.playlist.books else { return }
-            do {
-                let sortedBooks = try BookSortService.init(books: books).perform(filter: .fileName)
-                self.playlist.books = sortedBooks
-                DataManager.saveContext()
-                self.tableView.reloadData()
-            } catch {
-                print(error)
-            }
-
-        }))
-
-        alert.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: { (action) in
-            alert.dismiss(animated: true, completion: nil)
-        }))
-
-        present(alert, animated: true, completion: nil)
-    }
-    
     override var items: [LibraryItem] {
         return self.playlist.books?.array as? [LibraryItem] ?? []
     }
@@ -54,7 +19,7 @@ class PlaylistViewController: BaseListViewController {
         super.viewDidLoad()
 
         self.toggleEmptyStateView()
-
+        self.playlist.delegate = self
         self.navigationItem.title = playlist.title
     }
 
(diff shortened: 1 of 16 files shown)


GianniCarlo:
I glanced it over, and this looks very promising! I'll have more to say tomorrow👌
Thanks again @ryantstone for you continued contributions to the project 👍

ryantstone:
Is the app using any error reporting services? (Rollbar, instabug, etc) if so I should update the pr to report any sort errors.

GianniCarlo:
We don't actually use one of those services yet 🤔, we've got a simple contact support from the settings page. Although I do see that we could use one of those for easier bug reports. Would you mind creating an issue with what you think is our best option?

BookPlayer/Models/Playlist+CoreDataClass.swift
```diff
@@ -123,4 +126,18 @@ public class Playlist: LibraryItem {
         self.desc = "\(books.count) Files"
         self.addToBooks(NSOrderedSet(array: books))
     }
+
+    func sort(by sortType: PlayListSortOrder) {
```
GianniCarlo: let's add a `throws` to the function, so the error can be handled where it's being called, and we avoid declaring a delegate

BookPlayer/Services/BookSortService+SortError+PlayListSortOrder.swift
```diff
@@ -0,0 +1,42 @@
...
+
+final class BookSortService {
+
+    var books: NSOrderedSet
+
+    init(books: NSOrderedSet) {
+        self.books = books
+    }
+
+    public func perform(filter type: PlayListSortOrder) throws -> NSOrderedSet {
```
GianniCarlo: I don't think `perform(filter type...` conveys the appropriate meaning, since we're not actually filtering but sorting. we could rename it to `sort(type: PlayListSortOrder...`

BookPlayer/Services/BookSortService+SortError+PlayListSortOrder.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+final class BookSortService {
+
+    var books: NSOrderedSet
```
GianniCarlo: I'm still debating if this class needs a property `books` 🤔, since an instance of this class is actually only used immediately for the `perform` function and then never used again. Wouldn't it make more sense to make `perform` a class function, and pass the books as a parameter? thoughts?

BookPlayer/Library/PlaylistViewController.swift
```diff
@@ -96,6 +96,26 @@ class PlaylistViewController: BaseListViewController {
     @IBAction func addAction() {
         self.presentImportFilesAlert()
     }
+
+    @IBAction func didTapSort(_ sender: Any) {
+        let alert = UIAlertController(title: "Sort Files", message: "Sort Playlist files by", preferredStyle: .actionSheet)
```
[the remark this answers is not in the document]
pichfl: I'll have to take a look at this later this week, but at first glance I'd go with the title only. `Sort files by` should be sufficient.

BookPlayer/Services/BookSortService+SortError+PlayListSortOrder.swift
ryantstone: I agree. One of those situations where you start the class and put what you think you need in place and end up not needing it in the end.

BookPlayer/Models/Playlist+CoreDataClass.swift
ryantstone: 👍

ryantstone:
Suggest changes sound good to me. I'll work on when I get home this evening.

ryantstone:
Made the suggested changes.

BookPlayer/Services/BookSortService+SortError+PlayListSortOrder.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+
+final class BookSortService {
```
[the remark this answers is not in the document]
ryantstone: It could, but having sorting being handled by the model seems to be mixing concerns to me.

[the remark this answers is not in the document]
ryantstone: I was thinking about this on the way to work and another possibility is to pass a playlist to the service and this would remove the need to have a sort method in the playlist itself. Let dependency injection handle the whole thing.

pichfl:
@ryantstone Terribly sorry for the delay on my end. I think it works as expected. After testing on my device I have two remarks.

1. As Gianni already stated, I would love to make the Library sortable as well - I wonder if Library and Playlist should share a common ancestor or respond to the same protocol for this.
2. After spending some time with the interface I would suggest this for the dialog:

> **Sort files by**
> Title  
> Original filename

Again, sorry for making you wait and making you rehash the same things over and over.

ryantstone:
No worries. I’ve been under the weather over the weekend and I’m still recovering. Once I’m feeling better I’ll be able to take a look at this again. When is the release targeted for?

GianniCarlo:
@ryantstone we're thinking on releasing 3.2.0 by the end of the month if possible. We'll push watch and carplay support for 3.3.0 though

GianniCarlo:
🎉 thanks for the ongoing contributions @ryantstone , it's been a real big help for us

ryantstone:
No problem, at all. Anything to have a better audiobook player on the market. What is the next priority for the release?

PR #263, opened 2018-10-16
WIP: CarPlay

pichfl:
This also includes a refactored playback queue.

Do not merge yet, still missing a few things

- [ ] Playlists in CarPlay should have an additional first item "Continue" with appropriate icon
- [ ] Playing item should have an "now playing" indicator (Apple Music uses a loudspeaker)
- [x] Complete playback refactoring
- [ ] Update tests

### DIFF
--- a/BookPlayer/AppDelegate.swift
+++ b/BookPlayer/AppDelegate.swift
@@ -59,6 +59,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // register document's folder listener
         self.setupDocumentListener()
 
+        // register for CarPlay
+        self.setupCarPlay()
+
         return true
     }
 
@@ -218,6 +221,25 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             PlayerManager.shared.rewind()
             return .success
         }
+
+        //
+        /*
+        let speedOptions = PlayerManager.speedOptions.reversed().map({ (option) -> NSNumber in
+            return NSNumber(value: option)
+        })
+
+        MPRemoteCommandCenter.shared().changePlaybackRateCommand.supportedPlaybackRates = speedOptions
+
+        MPRemoteCommandCenter.shared().changePlaybackRateCommand.addTarget { (commandEvent) -> MPRemoteCommandHandlerStatus in
+            guard let cmd = commandEvent as? MPChangePlaybackRateCommandEvent else {
+                return .success
+            }
+
+            PlayerManager.shared.speed = cmd.playbackRate
+
+            return .success
+        }
+         */
     }
 
     func setupDocumentListener() {
@@ -225,5 +247,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         self.watcher = DirectoryWatcher.watch(documentsUrl) {
             DataManager.notifyPendingFiles()
         }
+    }
+
+    func setupCarPlay() {
+        MPPlayableContentManager.shared().dataSource = CarPlayManager.shared
+        MPPlayableContentManager.shared().delegate = CarPlayManager.shared
     }
 }
--- a/BookPlayer/Library/LibraryViewController.swift
+++ b/BookPlayer/Library/LibraryViewController.swift
@@ -31,11 +31,10 @@ class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate
 
         var books = [Book]()
 
-        if let playlist = item as? Playlist,
-            let index = playlist.itemIndex(with: identifier) {
+        if let playlist = item as? Playlist, let index = playlist.itemIndex(with: identifier) {
             books = playlist.getBooks(from: index)
         } else if let lastPlayedBook = item as? Book {
-            books = self.queueBooksForPlayback(lastPlayedBook)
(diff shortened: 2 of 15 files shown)


BookPlayer/AppDelegate.swift
```diff
@@ -59,6 +59,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // register document's folder listener
         self.setupDocumentListener()
 
+        // register for CarPlay
+        self.setupCarPlay()
```
GianniCarlo: this should be called inside: https://developer.apple.com/documentation/carplay/cpapplicationdelegate/2968287-application

GianniCarlo:
as discussed, we should have some enum that helps us keep track with how carplay manages indexPaths

BookPlayer/Player/PlayerManager.swift
```diff
@@ -54,33 +52,42 @@ class PlayerManager: NSObject {
...
+                        self.jumpTo(book.currentTime)
+                    }
+
+                    // Set speed for player
+                    audioplayer.rate = self.speed
+
+                    NotificationCenter.default.post(name: .bookReady, object: nil, userInfo: ["book": book])
+
+                    if #available(iOS 11.0, *) {
+                        MPNowPlayingInfoCenter.default().playbackState = self.isPlaying ? MPNowPlayingPlaybackState.playing : MPNowPlayingPlaybackState.paused
```
GianniCarlo: is this needed? only asking since the [doc](https://developer.apple.com/documentation/mediaplayer/mpnowplayinginfocenter/2588243-playbackstate) states that this only applies for MacOS 🤔

BookPlayer/Models/Playlist+CoreDataClass.swift
```diff
@@ -100,27 +115,43 @@ public class Playlist: LibraryItem {
...
+            guard let book = item as? Book, !book.isCompleted else { continue }
+
+            return book
+        }
+
+        return nil
+    }
+
+    func getNextBook(after book: Book) -> Book? {
+        guard let books = self.books?.array as? [Book] else { return nil }
```
GianniCarlo:
```suggestion
        guard 
          let books = self.books?.array as? [Book],
          let index = books.firstIndex(of: book),
          index + 1 <= books.count else { 
            return nil 
        }
        
        return books[index + 1]
```
Trying out the new Github feature, although it doesn't have an option to let me affect more lines :/

BookPlayer/Library/LibraryViewController.swift
```diff
@@ -294,23 +298,24 @@ class LibraryViewController: BaseListViewController, UIGestureRecognizerDelegate
 }
 
 // MARK: - TableView Delegate
+
 extension LibraryViewController {
-    func tableView(_ tableView: UITableView, editActionsForRowAt indexPath: IndexPath) -> [UITableViewRowAction]? {
+    func tableView(_: UITableView, editActionsForRowAt indexPath: IndexPath) -> [UITableViewRowAction]? {
```
GianniCarlo: interesting 🤔 I didn't know you could just leave '_' w/o specifying the name of the parameter in the function declaration if it isn't being used

BookPlayer/Models/Playlist+CoreDataClass.swift
pichfl: I'm not sure if I still like squeezing everything in a single guard just because it's possible. I did before, but I think it gets very hard to read. Maybe this requires further discussion elsewhere.

BookPlayer/AppDelegate.swift
pichfl: That method is only available in iOS 12. Should we consider raising the target?

GianniCarlo: mmmmm a week ago I think I read it reached 50% adoption rate. There were other reasons we were considering raising the target (I don't remember which ones now though 🤔). Honestly I lean towards raising the target

BookPlayer/Models/Playlist+CoreDataClass.swift
GianniCarlo: I see your point about readability, and in some cases it may make sense to split the guards. The root of my suggestion is changing the `if let` into a `guard`, to adhere to the [golden path](https://github.com/raywenderlich/swift-style-guide#golden-path), once I saw:
```swift
guard let books = self.books?.array as? [Book] else { return nil }

guard let index = books.firstIndex(of: book), index + 1 <= books.count else { return nil }

return books[index + 1]
```
my mind grouped the first two guards

pichfl:
After testing the current build in my car I found that I have a few more issues left.

Right now it is not possible to jump to the currently playing book by selecting it from the list, the app responds with an error message. Also I messed up the logic that presents the player inside the app while refactoring.

pichfl:
Will rebase asap

pichfl:
Actually no, I won't rebase this. As discussed with @GianniCarlo I'll move CarPlay to a separate release branch, keeping it out of the regular release for now, given the hassle this creates for developers without then necessary entitlements. I'm closing this PR for now.

Issue #265, opened 2018-10-25
Autoplay not working on files imported by airdrop


flo-assling:
Hey,

I love that App.
What I want to do with it, is play an Audiobook, which is split into many MP3‘s - 1 per chapter.
I transferred them via Airdrop, imported into one Playlist.
I enabled „Autoplay Library“, but playback stops after every MP3 (chapter)...

GianniCarlo:
hi @flo-assling , for testing purposes, can you please try out the autoplay with the screen on? 

In the latest release we included a fix for the autoplay not working when the iPhone was locked, since the files were being imported with the default file protection which is to lock access of everything when the iPhone was locked

flo-assling:
Yes, it works while the screen is unlocked :)

Does that mean that i only have to wait for an updated App?

GianniCarlo:
Thanks for reporting this @flo-assling, it seems that the fix we included doesn't cover all the cases, we'll do some testing with the steps you provided and we'll have a fix for the next release 👌

flo-assling:
Understood. I am on a iPhone Xs Plus with iOS 12.0.1

GianniCarlo:
closed by #267

Serendigity:
I am finding this same problem as of 13Nov18 using iOS latest version 12.0.1 with iPhone Xs Plus using Airdrop to transfer files.

pichfl:
@Serendigity what version of BookPlayer is installed on your device? You can find the precise version string inside Settings.

GianniCarlo:
thanks @Serendigity 👌

Serendigity:
I have been listening to an audiobook with the latest version of the app and did not notice any problem when moving from one file to the next. BookPlayer 3.2.0-9 on iOS 12.1.  I can confirm the issue is resolved for me.

GianniCarlo:
Thanks @Serendigity for letting us know

PR #266, opened 2018-10-27
Refactor autoplay

pichfl:
This should help making the autoplay more robust.

Re #265

### DIFF
--- a/BookPlayer/Models/Book+CoreDataClass.swift
+++ b/BookPlayer/Models/Book+CoreDataClass.swift
@@ -126,23 +126,17 @@ public class Book: LibraryItem {
             return next
         }
 
-        guard
-            UserDefaults.standard.bool(forKey: Constants.UserDefaults.autoplayEnabled.rawValue),
-            let library = self.library
-        else {
+        guard UserDefaults.standard.bool(forKey: Constants.UserDefaults.autoplayEnabled.rawValue) else {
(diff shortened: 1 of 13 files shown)


BookPlayer/Models/Book+CoreDataClass.swift
```diff
@@ -106,10 +106,45 @@ public class Book: LibraryItem {
...
     }
+
+    override func getBookToPlay() -> Book? {
+        return self
+    }
+
+    func nextBook() -> Book? {
+        if
+            let playlist = self.playlist,
+            let next = playlist.getNextBook(after: self) {
```
GianniCarlo: This is crashing with index out of range when it finishes the last item of a playlist

BookPlayer/Models/Book+CoreDataClass.swift
```diff
@@ -106,10 +106,45 @@ public class Book: LibraryItem {
...
+        }
+
+        guard
+            UserDefaults.standard.bool(forKey: Constants.UserDefaults.autoplayEnabled.rawValue),
+            let library = self.library
+        else {
+            return nil
+        }
+
+        let nextItem = library.getNextItem(after: self.playlist ?? self)
```
GianniCarlo: also crashing when library autoplay is enabled and the last item finishes playing

PR #267, opened 2018-10-29
Migrate CoreData stack + Autoplay fix

GianniCarlo:
In preparation for future features, this PR enables App Groups and migrates the existing CoreData stack. If for some reason the migration fails, the fallback is to load into the library all of the existing files in the `Processed` folder.

I'm also slipping in the fix for #265 since the last time I fixed it, I only changed the files permissions of the existing files and forgot to put that condition when importing new files 😅


GianniCarlo:
👌 I agree on the `.isFault` thing. We'll come back to it later

PR #268, opened 2018-10-31
Siri shortcut for last played book

GianniCarlo:
I'm adding the last played book shortcut using `NSUserActivity`

`UserActivityManager` holds one playback activity to resume or stop in tandem with the playback (supposedly this will help with siri suggestions too)

I wasn't sure about the `loadLastBook`. Since the app could be closed when the shortcut is triggered, the library needs to be loaded, that's why `DataManager.getLibrary()` was moved to `RootViewController` and why `loadLastBook` exists on `RootViewController. 
I wasn't sure about using the Notification center, since a notification could arrived when the `LibraryViewController` hasn't loaded yet, so the observers aren't set yet.

I'd like feedback on this last part, I'm still not convinced if the way I'm doing it is the best way

### DIFF
--- a/BookPlayer/Constants.swift
+++ b/BookPlayer/Constants.swift
@@ -37,7 +37,5 @@ enum Constants {
         case boosted = 2.0
     }
 
-    enum Activities: String {
-        case playback = "com.tortugapower.audiobookplayer.activity.playback"
-    }
+    static let UserActivityPlayback = Bundle.main.bundleIdentifier! + ".activity.playback"
 }
--- a/BookPlayer/Constants.swift
+++ b/BookPlayer/Constants.swift
@@ -36,4 +36,8 @@ enum Constants {
         case normal = 1.0
         case boosted = 2.0
     }
+
+    enum Activities: String {
+        case playback = "com.tortugapower.audiobookplayer.activity.playback"
+    }
 }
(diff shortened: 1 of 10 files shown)


BookPlayer/Constants.swift
```diff
@@ -36,4 +36,8 @@ enum Constants {
         case normal = 1.0
         case boosted = 2.0
     }
+
+    enum Activities: String {
+        case playback = "com.tortugapower.audiobookplayer.activity.playback"
+    }
```
ryantstone: Should this & the `info.plist` file info be dynamic to reflect the BundleID so we don't have a mix of identifiers for forked versions?

`case playback = "\(NSBundle.mainBundle().bundleIdentifier).activity.playback"`

GianniCarlo: totally, thanks for pointing this out @ryantstone 👌

ryantstone: It might be reasonable with the number of us working on the project to at some point in the near future move to a method of handling Bundle ID and possibly provisioning profiles to another setup so we're not accidentally including our provisioning profiles and custom ID's in PR's

GianniCarlo:
updated with your suggestion @ryantstone , also updated the PR to avoid changing stuff in the `RootViewController`, in the end it's best check the `options` in `didFinishLaunchingWithOptions` to store a boolean in the UserDefaults, so when the last book is loaded at the end in the `LibraryViewController`, it only has to check if it should play the book if the key exists.

GianniCarlo:
I modified the app groups identifier too, it now uses the bundle identifier macro, (if you've already run the branch `develop`, running this one will clear your library since the migration won't trigger a second time

PR #275, opened 2018-11-09
App Store Release 3.2.0

GianniCarlo:
We're going live with:
- Siri shortcuts (last played book)
- Url scheme (bookplayer://), only action available is play last played book
- Autoplay fixed (again)
- Sort button for library and playlists
- Zip files support
- Clarified chapter subtitle in linst
- Export/Share book option

Thanks to @pichfl , @ryantstone , @atomicguy for the PRs and also to the users giving us feedback and making 3.2.0 possible 🎉


GianniCarlo:
live on the app store 🎉

PR #276, opened 2018-11-10
Completed state

ryantstone:
- Adds Completion state attribute to `LibraryItem`
- Adds `UITableViewRowAction` to `LibraryViewController` (the color is not good for the button, need a suggestion for a better option)
- Adds `UIAlertAction` for marking completion state of playlist
  - This will mark all child books in a playlist as complete


pichfl:
I think it's time we move all actions into a single button that brings up a popup with all options, if possible having the swipe over action bringing up the sheet directly. This way we can unify the interaction in both Library and playlists while also having enough space for everything else.

ryantstone:
Agreed. I also think we could benefit from moving that logic out of the main view controllers into a subclassed alert(s) because there is a signification portion of `PlayListViewController` and `LibraryViewController` devoted to logic handling of alertss.

pichfl:
Unless @GianniCarlo objects, I think this could work quite well and reduce both controllers to something more manageable.

GianniCarlo:
I agree, it's been in the back of my mind for some time now to relocate those things too 👍

GianniCarlo:
I'll test this out between today and tomorrow and leave my feedback 👍

BookPlayer/Library/BaseListViewController.swift
```diff
@@ -130,8 +130,15 @@ class BaseListViewController: UIViewController {
         }
     }
 
+    func setupPlayer(books: [Book] = []) {
```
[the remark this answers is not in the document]
ryantstone: I think this was a goof on my part during a rebase. I saw it when I was reviewing the PR and was trying to figure out where it came from but didn't have time to dig into at that moment.

BookPlayer/Player/PlayerManager.swift
```diff
@@ -407,15 +407,20 @@ extension PlayerManager: AVAudioPlayerDelegate {
...
 
-        player.currentTime = player.duration
-
         UserDefaults.standard.removeObject(forKey: Constants.UserDefaults.lastPlayedBook.rawValue)
 
+        self.currentBook?.setCompletionState()
+
+        self.currentBook?.resetProgress()
+
+        self.pause()
```
GianniCarlo: I don't think this `pause` is necessary here either, can't see the difference when I remove it

BookPlayer/Models/Playlist+CoreDataClass.swift
```diff
@@ -164,6 +163,13 @@ public class Playlist: LibraryItem {
 
         return "\(count) Files"
     }
+
+    public override func setCompletionState(isComplete: Bool = true) {
+        books?.forEach { book in
+            guard let book = book as? Book else { return }
+            book.setCompletionState()
```
GianniCarlo: this should receive the parameter `isComplete` that is being passed in the function, currently if I set a playlist to not complete, it will still mark the books as completed

ryantstone:
Haven't forgotten this, just been busy. Hope to get to this next week some time.

GianniCarlo:
hey @ryantstone I just noticed that [1159708](/TortugaPower/BookPlayer/pull/276/commits/1159708e2b6474ef6b6fdbbefb14727e1693171e) was forced pushed some time ago, let us know if that's the final commit so we can review again, or if you're working on it some more

ryantstone:
I'll try to take a look this weekend. I recently had a kid so I haven't had any free time to sit at a computer. If I remember correctly I had 1 outstanding thing on this I just don't recall what it was.

GianniCarlo:
oh wow O: don't worry about it, family first 👍. I'll comment then on your latest commit so you can look back when you have the time.

GianniCarlo:
let me know if you'd rather I take over this PR, my brother has had a kid for about 1 year now, and he's always exhausted 😅

GianniCarlo:
Ryan I'm taking over this in #321 , congratulations on having a kid 😁, and we'll be right here whenever you get some free time in your hands

PR #278, opened 2018-11-12
Add an option to always use dark theme

micthiesen:
Thoughts on this feature? I'd find it useful, especially late at night. Also, this is my first time using Swift and/or iOS development so I've probably done something wrong 🙂

### DIFF
--- a/BookPlayer/AppDelegate.swift
+++ b/BookPlayer/AppDelegate.swift
@@ -27,7 +27,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             defaults.set(true, forKey: Constants.UserDefaults.chapterContextEnabled.rawValue)
             defaults.set(true, forKey: Constants.UserDefaults.smartRewindEnabled.rawValue)
             defaults.set(true, forKey: Constants.UserDefaults.completedFirstLaunch.rawValue)
-            defaults.set(true, forKey: Constants.UserDefaults.alwaysDarkThemeEnabled.rawValue)
+            defaults.set(false, forKey: Constants.UserDefaults.darkThemeEnabled.rawValue)
         }
 
         // Appearance
--- a/BookPlayer/AppDelegate.swift
+++ b/BookPlayer/AppDelegate.swift
@@ -27,6 +27,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             defaults.set(true, forKey: Constants.UserDefaults.chapterContextEnabled.rawValue)
             defaults.set(true, forKey: Constants.UserDefaults.smartRewindEnabled.rawValue)
             defaults.set(true, forKey: Constants.UserDefaults.completedFirstLaunch.rawValue)
+            defaults.set(true, forKey: Constants.UserDefaults.alwaysDarkThemeEnabled.rawValue)
         }
 
         // Appearance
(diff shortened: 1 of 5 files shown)


GianniCarlo:
>Thoughts on this feature? I'd find it useful, especially late at night. Also, this is my first time using Swift and/or iOS development so I've probably done something wrong 🙂

This is great!, I'd like to expand this to be the setting for dark mode for the app though if you're up for it @micthiesen , I'm sure that @pichfl can weigh in with the color scheme if that's the case 🤔

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

micthiesen:
Sure I can give it a go! I think an app-wide dark theme would be great. It'll be a challenge (for me at least) so I can't promise I'll be successful.

micthiesen:
Also I just noticed this PR has some problems: Using a dark these for books that do not contain artwork, and refreshing the player when the settings toggle changes. I think fixing these will require a larger refactor than what I've changed already.

GianniCarlo:
Good catch, and don't worry, we'll help you out along the way in whichever way we can. I'll be able to check this out more in depth later today and maybe give some suggestions 👌

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

GianniCarlo:
Alright to expand on this, we should make use of the [Notification Center](https://developer.apple.com/documentation/foundation/notificationcenter), we already use it to keep many of the things in the app up to date. So the steps would be:
- Create a new constant in the file `Notification+BookPlayer.swift` inside the `Extensions` folder
- Register the observer for the new notification in the `BaseListViewController` (the Library and Playlist ViewControllers inherit from this class)
- Also make the `PlayerViewController` register to this notification, there's already a function there (`setupView(book currentBook:Book)`), something that eventually calls this function may work to reset the player view
- Make the `SettingsViewController` post the notification when the switch is toggled

micthiesen:
@GianniCarlo thanks for the pointers! I'll give it a go and let you know if I find any roadblocks. I've created issue #279 to keep track of things.

pichfl:
II'm a bit in the water right now, but I'll try to whip up something for the library. Considering the fact that this will mostly shine on OLED displays, I suggest we use a full black for the background. 

I will get back to you with a set of screens and colors until the weekend.

pichfl:
I think I may have done something wrong there when I implemented the color extraction. When extracting the colors, we should extract two sets, both a light and a dark theme. Otherwise the cover art will favor whatever theme was active when the user imported the book.

micthiesen:
Ah that's how we should approach it. I was struggling with that. Thanks. I can refactor the code to extract both sets. Love the idea of using full black for OLEDs.

I'll also have to look into building a "migration" for existing books, since they'll only have the one set of colours. Ideally this us run whenever the colour extraction process changes (maybe whenever the app updates).

And no rush. This may take me a while to implement 😅

GianniCarlo:
@micthiesen thanks for getting things started, I took over and opened up a WIP PR #309  with the feature almost fleshed out 👌, you're welcome to try it out and give feedback as well.

GianniCarlo:
I'm closing this PR and moving further discussion to #309

micthiesen:
I am totally guilty of forgetting about this. Thanks for taking it on, I’m curious to see your solution.

Issue #224, opened 2018-07-23
Fine grained playback speed settings


bensaufley:
**Feature Request**: allow custom speed (within reason). I mostly use the Audible app right now, and it supports speeds similar to BookPlayer. Google Play's app, which I just tried out, allows increments of `.05` anywhere within a range, using a slider UI. It's great and solves one of my biggest problems—I've had a few books that are too slow for 1.5x, but too fast for 2x. Even just supporting 1.75x would be cool, but a custom speed slider would be 👌

I'd love to contribute if you think this is something doable. Swift isn't really my primary language, but I've done a little.

FWIW my environment is iOS 11.4.1 on an iPhone 7 Plus, running BookPlayer 3.0.0. Please let me know if there's anything else I can let you know.

pichfl:
Wonderful idea! Whilst not on our Roadmap yet, we discussed replacing the speed (& sleep timer) controls with a custom interface that allows for the fine adjustments you're looking for. 

If you want to tackle this yourself and need help getting into the project, feel free to contact me personally via Twitter or Email. We would be happy to get you involved. 

Either way, I'll provide interface mockups for this that match the overall look and feel of the new UI.

To keep our backlog clean and focussed, I'm closing this ticket and adding it to our roadmap.

pichfl:
<img src="https://user-images.githubusercontent.com/194641/53362790-3b2a6a80-393b-11e9-9013-b6843c0960a1.png" alt="" width="375">

This is how the new interface could look like. A slider which locks in 1/8 steps from 0.625 to 2.625. Tapping the speed icon or swiping down will dismiss the card. Tapping outside the card will do so as well. 

If someone want's to tackle this, please contact us so we can coordinate the necessary changes to be made to the overall player UI discussed in #281

bensaufley:
Just coming back to this. Not really able to contribute right now unfortunately (Swift not being really an area of expertise and time being limited). But I'd add that it'd be nice to go above 2.5x. I had to switch back to the Audible player for a book recently so that I could do 3x (some readers are just so slow!). Audible and Libro.fm both support 3x speed; Audible supports 3.5x, even. Can't support every use case, obviously, but worth considering the way to best support the broadest number of use cases without overcomplicating the UI and the code.

(Also I like your choice of example book with the mockup @pichfl)

pichfl:
[...]

PR #280, opened 2018-11-13
Playback Speed resolution Change

woodduckie:
Add additional granularity (1.1) for playback speed.  Often audiobooks can be played at a slight increase in speed, but 1.25 can be too rapid for clarity.

### DIFF
--- a/BookPlayer/Player/PlayerManager.swift
+++ b/BookPlayer/Player/PlayerManager.swift
@@ -15,7 +15,7 @@ import MediaPlayer
 class PlayerManager: NSObject {
     static let shared = PlayerManager()
 
-    static let speedOptions: [Float] = [2.5, 2, 1.5, 1.25, 1, 0.75]
+    static let speedOptions: [Float] = [2.5, 2, 1.5, 1.25, 1.1, 1, 0.75]
 
     private var audioPlayer: AVAudioPlayer?
 


GianniCarlo:
hey @woodduckie thanks for the contribution 👍. Although I would prefer that instead of adding more options (via support someone requested 1.75x also), we should have a 'Set Custom Speed' option or something like it. Upon selection, it could present an alert with a textfield so the user can input their preferred speed. we would have to put checks on it obviously.

Let me know what you think

woodduckie:
I overlooked the closed issue #224 - apologies for the redundant submission.  Feel free to close this PR & leave with the development roadmap.

I think a visual linear slider or rotary selection might be the most flexible & friendly solution as described, but beyond my skills at this point :)

A custom speed setting with text entry would be a good first step that would provide flexibility as desired without a larger rewrite of the interface.

GianniCarlo:
No problem, if you're up for it, let us know 👌 it'd be a great addition for v3.3.0

bryanrezende:
Another option instead of a text input is a dedicated “Playback Options” dialog of sorts, which could house any quick-access playback toggles/values. This is how Pocketcasts handles it: 

Image
Image

GianniCarlo:
that's a very good and slick alternative 👍

bryanrezende:
@pichfl @GianniCarlo I wouldn't mind tackling creating a playback options dialog/overlay like this if you like the approach, as a way to refamiliarize myself with iOS (been away and on web for nearly a year). I'm going to take some time to get familiar with the huge evolution of the codebase as well!

pichfl:
@bryanrezende I just discussed this with @GianniCarlo and I think we should do this properly which will require a few more things. I think it's worthwhile to discuss this in a separate issue so I created #281 

You can also join us on Discord if you want, we are slowly moving over from iMessage to make it easier to discuss with everyone involved.

PR #289, opened 2018-12-11
Batch edit and delete

GianniCarlo:
This replaces the 'Sort' button at the top right corner, in favor of a new button 'Edit', which when activated, will dismiss the mini player, and show new controls (Sort | Move | Trash).

In this PR, I've introduced some protocols that will provide default implementation for things. You can look at the diff of the last commit to see what I've moved around. I'd appreciate feedback specially on this topic.


GianniCarlo:
The protocol approach can be expanded, for the time being I created four initial protocols as a start
- ItemList
- itemListActions
- ItemListFeedback
- ItemListAlerts

Not sure about those last two though

Issue #284, opened 2018-11-15
Playlist not sorted properly


halleygen:
* Do you want to request a *feature* or report a *bug*?
Report a bug

* What is the current behavior?
I have a playlist of files titled "Chapter 1", "Chapter 2", "Chapter 3", etc. When I try to sort them using the new sort feature added in v3.2.0 they are sorted as: "Chapter 1", "Chapter 10", "Chapter 11", "Chapter 2", instead of the correct order (like in Finder or Files.app)
 
* If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo / screenshot / screen recording of the problem. 
![img_d4fec02a4d4b-1](https://user-images.githubusercontent.com/21304129/48584223-70312500-e929-11e8-94b7-fa6ff54201e0.jpeg)

* What is the expected behavior?
For the files to be ordered correctly, like in Finder or Files.app

* What is the motivation / use case for changing the behavior?
To be able to listen to audiobooks broken into multiple files in the correct order

* Please tell us about your environment:
    * Version of BookPlayer: v3.2.0
    * Your Device and its OS version: iPhone 6s, iOS 12.1

* Other information
I think this can be solved by using the `NSString.CompareOptions.numeric` option to compare filenames when sorting the files

ryantstone:
Are you sorting on filename or on title?

halleygen:
Filename, but I just tried and sorting by the title doesn’t work properly either.

ryantstone:
Ok. I’ll take a look. Thanks for pointing it out.

GianniCarlo:
I ran into this while I was refactoring some things, turns out we had to use [.numeric](https://developer.apple.com/documentation/foundation/nsstringcompareoptions/nsnumericsearch?language=objc) as an option for the `NSSortDescriptor`. The fix is included in the batch edit PR #289

GianniCarlo:
closed by #290

halleygen:
This works great in the new update. Thanks for you hard work!

PR #290, opened 2018-12-12
Fix sorting

GianniCarlo:
Extracted from #289. I also removed the `throws` from the protocol since the `sort` implementation doesn't throw, so the failure alert never would have been called


GianniCarlo:
solves #284

PR #292, opened 2018-12-17
Add option to sleep timer for 'End of chapter'

GianniCarlo:
This new option will stop current playback on the end of a chapter, or at the end of the book if it doesn't have any chapters (this is handy if the book is inside a playlist, it halts the autoplay)

~~Also I included a minor change in this PR, there was no defined behaviour for when the audio session failed to become active. For the time being, it'll crash on failing on activate, and it'll fail silently when it fails to deactivate. Reasoning behind this, there was a reliable way to make it fail to activate (don't remember now, I think it was with BookPlayer playing a book, start reproducing a video from within the reddit or twitter app, and trying to hit play on BookPlayer again), and it would 'start' but no sound would play, I had to force-close the app and launch it again to fix the audio session~~


GianniCarlo:
test fails due to the SwiftLint and the `force_try` rule in `try! AVAudioSession.sharedInstance().setActive(true)`. I'll remove the commit from the PR to make it only about the new sleep timer option, we'll discuss the fail-to-activate behaviour on Discord and see what would be the best option

GianniCarlo:
we'll close `3.2.1` with this PR

PR #309, opened 2019-01-25
Themes

GianniCarlo:
For those who want to have a dark theme before the next App Store release, you can use this branch.
This PR contains:
- DB migration to support the renaming of `ArtworkColors` into `Theme`
- Default Light and Dark theme
- A file named `Themes.json` that contains the extra themes to be added based on four colors
- Option to use a book's artwork as a theme

This PR was also rebased on #289 so it contains the batch edit feature as well

There's still work to be done on this feature, but I feel it's fleshed out enough that it's worth exposing it


GianniCarlo:
Preview:
img_1969trim

micthiesen:
This looks great! Is there still the ability to automatically set the theme based on what’s currently playing? I’m wondering if that option would be more useful than the manual extraction option.

GianniCarlo:
I like that idea a lot. I think they really complement each other (manual extraction should list the extracted theme among the available themes). We could throw it in as an option 🤔 as well

micthiesen:
Oh that’s a neat idea, simply creating a new theme based on manual extraction. I like it 👌

GianniCarlo:
After sinking too much of our time into making artwork colors viable to extract a theme from and failed. We decided that we'll remove it for the time being, and make the `PlayerViewController` follow the selected theme

PR #317, opened 2019-02-19
Fixed a cancelled `on chapter end` sleep timer not clearing properly.

taylorzane:
I wasn't sure if I should have opened an issue before opening this PR. If you'd like me to do so, let me know and I can cross-reference the issue and this PR.

This PR resolves an issue where a sleep timer set to "End of current chapter" is not cleared when selecting "Off". This doesn't appear to affect a properly chaptered book, but a playlist (after a quick overview of the code) appears to be treated as a collection separate books.


GianniCarlo:
@taylorzane good catch! 
Regarding creating an issue, since you already solved it here, it wouldn't be necessary. Before merging, could you point this PR to the branch `develop`, all our work for the next version is currently being done there.

Thanks for the contribution 😁

taylorzane:
I figured it was a simple enough change that it didn't warrant an issue anyway. I've updated the branch to be `develop` on your side.

I'm happy to help though! Thoroughly enjoying this app, so it's the least I could do. I'm hoping to contribute a bit more once the next version comes out; seems like you're pretty deep in a redesign so it's hard to really know the vision of things as an outsider. Let me know if there's anything I can look at to help out. 🙂

GianniCarlo:
that's great to hear! we'll definitely keep you in the loop (the next version is taking longer than planned, but we should be finishing up sooner rather than later)

Issue #314, opened 2019-02-07
Different app icons


GianniCarlo:
We'll include a way for the user to change app icons

pichfl:
Early draft for this view
<img src="https://user-images.githubusercontent.com/194641/52430179-d501b300-2b05-11e9-8270-bc9815d4787d.png" width="375">

I think I want to include at least one icon per theme so I'll be go creating some.

PR #318, opened 2019-02-19
App icons

GianniCarlo:
We brought back the v1 app icon!

This should be merged after the themes PR


GianniCarlo:
re #314

GianniCarlo:
To add new icons, they must be specified in both the app's `plist` and the `Icons.json` file. This file expects an array of objects that have these properties:
```
{
"id": <String>,
"title": <String>,
"artist": <String?>,
"imageName": <String>
}
```

Issue #312, opened 2019-02-05
First chapter starts when tap on the image of the playlist in the Library list


hydro0:
* Do you want to request a *feature* or report a *bug*?
Bug

* What is the current behavior?
When I press on the image of the book in the library it brings me a player with a first chapter

* If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo / screenshot / screen recording of the problem. 
Open some book on any chapter, then open library and press on the book image - it starts the first chapter. Can share the video if that’s not reproducible, but later when I will get to the laptop.

* What is the expected behavior?
It should open player view with current state.

* What is the motivation / use case for changing the behavior?
It loses entire progress and start the book from the first chapter, don’t think it is intentionally

* Please tell us about your environment:
    * Version of BookPlayer 3.2.1
    * Your Device and its OS version IPhone 7+, 12.2
* Other information
  Can help with this as well if you need spare hands

GianniCarlo:
It'd be of great help if you could provide a recording 👍, I'll be able to do some tests much later today and report back

hydro0:
Sorry, just read it opens first **unfinished** chapter, and somehow I had chapter 1 just a few seconds behind being finished. Though continuing from the current state looks more logical for me, but that's not a big deal, thanks for the fast reply!

hydro0:
Seems it is still reproducible even if I marked all chapters as read explicitly, please check the [video](https://share.icloud.com/photos/0h0CyieTjH36HqJ_FM5tyuA4A) :

1. All the files are "read" up to chapter 21.

1. When going back to the library and pressing on the artwork - chapter 1 is opened.

1. Marking chapter 1 as read again - just to be sure.

1. Open chapter 21.

1. Going back to the library and pressing on the artwork - chapter 1 is still opened.

pichfl:
Very important question: 
The chapters you're talking about, are they inside a single file or is *each file* a single chapter? The former would have files A-C with A containing Chapter 1-10, B containing Chapter 11-20, C containing 21-26, the latter having files A-Z with A being Chapter 1, and so on.

@GianniCarlo I think this may be solved by #276. I'm rather sure the error would be in this part of the model, which was my naive approach to finding the first unfinished file.
https://github.com/TortugaPower/BookPlayer/blob/6be02e1b427ee9e5a6620821fd35605d586fcd0b/BookPlayer/Models/Playlist%2BCoreDataClass.swift#L132-L142

[...]

Issue #201, opened 2018-07-10
Add `completed` boolean state for a book


pichfl:
Tracking via the duration has caused issues while working on 3.0 which are solved but maybe it's better to have this around.

GianniCarlo:
the 'pie' progress should be replaced with a checkmark, it sometimes does and sometimes doesn't. Currently we have a computed-value for `isCompleted`, and although we round off the `currentTime` vs `duration` it obviously isn't perfect, having a boolean as a core data property should be the better option

GianniCarlo:
we could set it to `true` when the `AVAudioPlayer` delegate calls the `audioPlayerDidFinishPlaying`

GianniCarlo:
also when it's explicitly marked by the user inside the player 'more' options

pichfl:
The reason the checkmark rarely shows up is that I'm still struggling with typed languages and comparing two floats is a bad idea. I put a preliminary fix into #266, casting the values to `Int` before comparing them.

bryanrezende:
I've been away from this project for ages now, but for what it's worth - you may not have to compare floats for equivalency to mark a book as complete. A book that's within (let's say) a minute of completion is likely complete due to the legal fluff that fills the end of a lot of books. The audible app counts things as complete as long as they are close enough to the end if I remember correctly, because at that point the book is functionally finished for the listener. 

A completed flag might still be nice to have though - if you wanted to record a completion date for whatever reason (stat/listening tracking?), the presence of a nonnull completion date could also function as a flag?

pichfl:
@bryanrezende welcome back 🙂 your remark is true for Audible as they strictly use M4B containers, but it doesn't work for MP3 based books which can be split into parts of 5 minutes or less and we have to account for those as well, as not all our users are willing to convert their books.

ryantstone:
I've got this started but have a few questions. We have a few ways to handle completion.

1. Detect when a book is close to done and when a user pauses present a dialog asking if they are done within 1-2 minutes (if it's a book that's split we can write some logic surround this functionality to overcome the 5 minute files)
2. Add a `UITableViewRowAction` for marking complete
3. When time is complete mark complete

Also, when adding this I went ahead and added `isComplete` to `LibraryItem` so this will encompass playlists too and we can have a callback when a book within a playlist completes and once all books are complete in a playlist update it's state too.

pichfl:
I don't think the dialog is necessary. I agree with everything else.

[...]

PR #321, opened 2019-02-25
Completed state & selected book on artwork tap

GianniCarlo:
- New `isFinished` boolean property for `LibraryItem`
  - This state is no longer tied to the progress of the book, so a book that is halfway through could be marked as finished and retain the current position that the user left it in.
- When tapping the artwork of a playlist, and the player has loaded a book of such playlist, it will open the player's book. This will prevent an 'earlier' book being loaded


GianniCarlo:
re: #312 #201

Issue #304, opened 2019-01-07
Move "Jump to start" and "Mark as finished"


cuddawudda:
from ... on bottom right of now playing screen to the library screen options. 

Personally I don't think such controls need to be in now playing screen as they are rarely used.

pichfl:
Personally, I don't agree. Why not have it in both places? Sometimes you're in the middle of a book when you decide it's not worth it or you're done. Actions should be as close to the item they apply to.

cuddawudda:
Hi both, nothin wrong with a difference of opinion!

For the moment there are not a lot of alternative functions that could be put there instead, but could i suggest that in the future there may be some functions that are more relevant to the now playing screen.
To me, it feels like library options such as these should be managed in the library view.

GianniCarlo:
The general mindset I had about this in particular, was to have in the library view for each cell: 
| Book cell (swiped left) | Options | Delete |
The 'Options' would present all the actions that are available in the '...' section of the player, as a way to access them without having to load the book or access the player

PR #322, opened 2019-02-25
Standardize available actions

GianniCarlo:
Now the cells have a single row action: 'Options', which will bring up an action sheet with all the available actions.

Standard actions for all:
- Rename
- Move
- Jump to Start
- Mark as Finished
- Delete

Special action for books:
- Export

This should be merged after #321


GianniCarlo:
Re #304

GianniCarlo:
sounds good, will update

PR #333, opened 2019-03-21
Setup Fastlane

GianniCarlo:


GianniCarlo:
after trying to fix things with Travis, and going through the issues and the documentation Fastlane provides, I no longer am thrilled about integrating Fastlane (https://github.com/fastlane/fastlane/issues/14364)

PR #342, opened 2019-04-05
Fix multiple bugs of 3.3.0

GianniCarlo:
I bundled the fixes to some crashes that were reported via sentry.io


GianniCarlo:
re #343

PR #353, opened 2019-04-23
"Always On" mode for when device is plugged in

bcongdon:
This PR adds a setting to allow disabling autolock when the device is plugged in to power.

I added this as a switch in the Settings view. If the "Disable Autolock" switch is off, then this setting will be inaccessible. (i.e. the user will need to enable "Disable Autolock" and then further enable "Only When Plugged In")

I also added a listener to battery state updates in the `PlayerViewController` so that the idle timer is enabled/disabled dynamically as the device battery state changes.

Please let me know if you'd prefer a different wording of the setting. This is my first iOS/Swift PR, so I'm also super open to other code/style suggestions. 😄

### DIFF
--- a/BookPlayer/Player/PlayerViewController.swift
+++ b/BookPlayer/Player/PlayerViewController.swift
@@ -110,17 +110,17 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
     }
 
     override func viewDidAppear(_ animated: Bool) {
-        if UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabled.rawValue) {
-            UIApplication.shared.isIdleTimerDisabled = true
-        }
+        self.updateAutolock()
     }
 
     override func viewWillDisappear(_ animated: Bool) {
         UIApplication.shared.isIdleTimerDisabled = false
+        UIDevice.current.isBatteryMonitoringEnabled = false
     }
 
     deinit {
         UIApplication.shared.isIdleTimerDisabled = false
+        UIDevice.current.isBatteryMonitoringEnabled = false
     }
 
     func setupView(book currentBook: Book) {
@@ -352,6 +352,26 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
     override func accessibilityPerformEscape() -> Bool {
         self.dismissPlayer()
         return true
+    }
+
+    func updateAutolock() {
+        guard UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabled.rawValue) else { return }
+
+        guard UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabledOnlyWhenPowered.rawValue) else {
+            UIApplication.shared.isIdleTimerDisabled = true
+            return
+        }
+
+        if !UIDevice.current.isBatteryMonitoringEnabled {
+            UIDevice.current.isBatteryMonitoringEnabled = true
+            NotificationCenter.default.addObserver(self, selector: #selector(self.onDeviceBatteryStateDidChange), name: UIDevice.batteryStateDidChangeNotification, object: nil)
+        }
+
+        UIApplication.shared.isIdleTimerDisabled = UIDevice.current.batteryState != .unplugged
+    }
+
+    @objc func onDeviceBatteryStateDidChange() {
+        self.updateAutolock()
     }
 }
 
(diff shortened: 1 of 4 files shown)


GianniCarlo:
Hi @bcongdon ! This looks great 👍, I'll be able to test this out later today and have more comments, in the mean time, please make this PR point to our branch `develop`

Ps. thanks for your contribution 😄

pichfl:
Thanks for joining our efforts, this is much appreciated 👍 

Small wording change from my side: `Only when connected to power`. We try to only capitalize words that are used as names such as "Library" meaning the library of books inside BookPlayer, and Autolock being a feature of the OS. I think it could be more consistent (your PR makes a strong case for lacking that), but at least that's what we were aiming for.

bcongdon:
Sounds good! I updated the label to use the capitalization you described

BookPlayer/Player/PlayerViewController.swift
```diff
@@ -110,17 +110,19 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
     }
 
     override func viewDidAppear(_ animated: Bool) {
-        if UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabled.rawValue) {
-            UIApplication.shared.isIdleTimerDisabled = true
-        }
+        UIDevice.current.isBatteryMonitoringEnabled = true
```
GianniCarlo: This is being enabled regardless if the setting is on or off. We should check that to avoid listening when the setting is off.

BookPlayer/Player/PlayerViewController.swift
```diff
@@ -353,6 +355,17 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
         self.dismissPlayer()
         return true
     }
+
+    func updateAutolock() {
+        let autolockDisabled = UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabled.rawValue)
+        let autolockDisabledOnlyWhenPowered = UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabledOnlyWhenPowered.rawValue)
+        let charging = UIDevice.current.batteryState == .full || UIDevice.current.batteryState == .charging
```
GianniCarlo: instead of going by `.full` and `.charging` we could check only for the `.unplugged` state

BookPlayer/Player/PlayerViewController.swift
```diff
@@ -110,17 +110,19 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
     }
 
     override func viewDidAppear(_ animated: Bool) {
-        if UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabled.rawValue) {
-            UIApplication.shared.isIdleTimerDisabled = true
-        }
+        UIDevice.current.isBatteryMonitoringEnabled = true
+        NotificationCenter.default.addObserver(self, selector: #selector(self.onDeviceBatteryStateDidChange), name: UIDevice.batteryStateDidChangeNotification, object: nil)
```
GianniCarlo: We can put this in `updateAutolock` and use the `UIDevice.current.isBatteryMonitoringEnabled` flag to know if the observer should be added

BookPlayer/Player/PlayerViewController.swift
```diff
@@ -353,6 +355,17 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
         self.dismissPlayer()
         return true
     }
+
+    func updateAutolock() {
```
GianniCarlo: after all the comments, this function would look something like:
```swift
guard UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabled.rawValue) else { return }

guard UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabledOnlyWhenPowered.rawValue) else {
  UIApplication.shared.isIdleTimerDisabled = true
  return
}

if !UIDevice.current.isBatteryMonitoringEnabled {
  UIDevice.current.isBatteryMonitoringEnabled = true
  NotificationCenter.default.addObserver(self, selector: #selector(self.onDeviceBatteryStateDidChange), name: UIDevice.batteryStateDidChangeNotification, object: nil)
}

UIApplication.shared.isIdleTimerDisabled = UIDevice.current.batteryState != .unplugged
```

BookPlayer/Player/PlayerViewController.swift
bcongdon: I was slightly concerned about the [`.unknown` battery state](https://developer.apple.com/documentation/uikit/uidevice/batterystate/unknown) so chose to explicitly check for `.full`/`.charging`. I'm not sure how common it is for the battery state to be unknown...

But it definitely is cleaner to just check for `!= .unplugged`. 👍

bcongdon:
Thanks for the feedback. 😄 I made the suggested changes

GianniCarlo:
Thanks again for your contribution @bcongdon 🙂

Issue #287, opened 2018-11-24
Sync between devices


pichfl:
I think using iCloud to sync playback position and the tiny amount of Library data between devices would be much welcome. Especially when the iPad version in #230 comes along.

Since we already support dead simple file sharing since #66 was added in 3.2.0 we don't have to sync files at all. While some people will find that too much manual work, I would argue that having fine grained manual control over the synced files without additional interface is enough of a advantage. We could also introduce a sharing option to the batch edit, allowing people to select their whole library for sharing/sync.

The most important thing for me would be proper sync of the playback position, as fast and accurate as possible.

GianniCarlo:
Agreed, this would be the next step, iCloud sync to provide an audible-like experience of having the metadata of your library/account synced to all your devices. After this one we'll figure out what to do with the files

dontcrash:
Any update :)?

GianniCarlo:
this is the main focus for BookPlayer in 2020, problem is, I haven't had much free time these last few months to put into BookPlayer, so it's slooowly moving forward :/

dontcrash:
> this is the main focus for BookPlayer in 2020, problem is, I haven't had much free time these last few months to put into BookPlayer, so it's slooowly moving forward :/

I haven’t had a chance to look over the code yet, but in some instances it is almost as simple as replacing NSPersistentContainer  NSPersistentCloudKitContainer (in AppDelegate) and then minor setup on the Apple Developer site

GianniCarlo:
I haven checked yet about CloudKit tbh, we were first reworking how we process our files, to have a 1:1 representation of the imported folder structure instead of what we're doing right now.

But if it's as simple as you put it, I'll give it a go 🤔

GianniCarlo:
An update on this issue, we did try CloudKit, and while syncing the CoreData database works automagically most of the time, moving the audio files would still need to be a manual process, otherwise we would need to use the user's storage space in iCloud for BookPlayer, possibly taking up double space for those that use iCloud as their cloud provider. For tech-savvy users, this won't be a problem, but for the rest of the users, it would be very confusing not being able to continue playback on their other devices due to files requiring to be imported (via airdrop maybe) from their main device

[...]

PR #357, opened 2019-05-01
Watch app support v1

GianniCarlo:
Finally watch support is coming.

This V1 will only act as a remote control for the main app. We'll leave for V2, things like syncing files and reproducing audio w/o the need of the phone being near.

The app consists of two pages:
- On the first page, you can navigate through your library and playlists, and see your last played book on top.
IMG_3975

- On the second page, we use the baked-in [Now Playing View](https://developer.apple.com/documentation/watchkit/adding_a_now_playing_view) which will show whatever was last played by the phone (it could be other apps, hence the need for the last-played-book on the library)
IMG_3974

Each time the app (phone) is launched, the application context is updated and the library information is synced to the watch. We also provide a menu (accessed by 3d-touch) on the watch with the option to manually request the phone to sync the library information. This could be improved, so no manual intervention is needed, but I think it will be best revisited for v2
IMG_3977

Things left to do:
- [x] Add empty screen state for the library
- [x] Add complication images


GianniCarlo:
The now-playing-view is buggy. At first launch it'll show the information correctly, but after some time, it's presented like this (shifted downwards) and then it won't work
IMG_4020

The only option left, is to build a custom player view and not rely on Apple's component

pichfl:
Whelp, only a zillion assets left to create 😅

GianniCarlo:
We got around Apple's bug using [reloadRootPageControllers](https://developer.apple.com/documentation/watchkit/wkinterfacecontroller/2868441-reloadrootpagecontrollerswithnam?language=objc) on each selection. It's not ideal in the long run, and it will be replaced with a custom player for v2

davidlist:
Hi, whats the status of v2? I think no other app provides this functionality and was the reason I discovered this project. Please let me know in case I can help! I am not a developer but I could help with testing or localization to German...

d1rewolf:
Agreed...this app would be perfect if it could sync files and play them on the watch without the phone...

GianniCarlo:
I'm currently focused on #287 , but if anyone in the community wants to contribute, please feel free to 😊 it'd be a huge help

davidlist:
I think there is still no other app to offer that functionality and that’s what brought me here. I would be willing to put time an effort into this however I am not a developer and my programming experience is very little. However if someone would want to take the lead and start with the basic constrict i would try to help es much as possible.

PR #365, opened 2019-06-13
Fixing issue with book not starting from the right time

gpambrozio:
When using the Siri intent AND the app is not in the background when the app starts it plays from the beginning of the previous book but not form where you played last.

This creates a better flag to check if the book should start playing or if it should postpone.

To reproduce: 

* Open the app
* Start a book and go to the middle of the file.
* Stop
* Background
* Hard quit
* Invoke the siri shortcut
* Book should start from the beginning.

The key to this change seems to be having a separate flag that indicates if the book was loaded, not if the audio player was created.

### DIFF
--- a/BookPlayer/AppDelegate.swift
+++ b/BookPlayer/AppDelegate.swift
@@ -70,7 +70,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if let activityDictionary = launchOptions?[.userActivityDictionary] as? [UIApplication.LaunchOptionsKey: Any],
             let activityType = activityDictionary[.userActivityType] as? String,
             activityType == Constants.UserActivityPlayback {
-            defaults.set(true, forKey: activityType)
+            self.playLastBook()
         }
 
         // Create a Sentry client and start crash handler
@@ -119,7 +119,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
-        PlayerManager.shared.play()
+        self.playLastBook()
         return true
     }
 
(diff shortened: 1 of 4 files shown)


GianniCarlo:
I think I may be doing something wrong, I'm attaching a gif with my understanding of the steps:

IMG_4905

Also, when `stop()` is called on the `PlayerManager`, this new variable should be set to `false`

gpambrozio:
Looks like what I did but this is what I get on latest develop branch (fe16bc54)

E619E0F9-FB78-4898-A8C3-4F2EF817E23A

gpambrozio:
What is the version of your ios? Mine is 12.2

GianniCarlo:
I'm on 12.3.1 iPhone 7. Mmm well regardless, if that fixes the problem then it's good 🙂. My only comment would be to see if we can remove the isLoaded property altogether in favor of the new one. And don't forget to update it to false on the stop() action

gpambrozio:
Maybe it's because I'm playing from a playlist and you're playing from a single file?

GianniCarlo:
I tried it with the book inside a playlist, but I got the same result 🤔, maybe OS version or older phone has something to do with it not happening?. Anyways, let's get rid of the ocurrences of `PlayerManager.shared.isLoaded` in favor of the new `PlayerManager.shared.hasLoadedBook` everywhere else, and set `hasLoadedBook` to false when the func `stop()` is called on the player (I think it only occurs when the mini player is dismissed manually).

And thanks again for your contribution @gpambrozio 🙂

gpambrozio:
I tried removing `isLoaded` but then it doesn't show the mini player when you start the app. I did reset the new flag in `stop` but I think removing `isLoaded` might be more complicated.

GianniCarlo:
ok I found out why the mini player isn't being presented, the line `self.hasLoadedBook = true` in `PlayerManager` should be set before the completion callback (inside the callback, the event to present the miniplayer is posted, so if the flag isn't set before the completion it won't work properly)

Screen Shot 2019-06-17 at 5 36 57 PM

gpambrozio:
That worked. I just didn't test the check in the AppDelegate's `showPlayer` that seems to originate in `application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:])`

GianniCarlo:
No worries, that func should be called when the app is loaded in memore either in background or foreground, so it should work 👍. Thanks again for your contribution @gpambrozio 🙂

Issue #369, opened 2019-08-14
Apple Watch App Does Not Show Clock


Serendigity:
With the new Apple Watch support there is no clock displaying the current time on the user interface. Please look at the layout of the default Apple 'Now Playing' app. 
![IMG_8817](https://user-images.githubusercontent.com/45026713/63003514-77389400-beb7-11e9-895d-04cfd7e354e1.JPG)
![IMG_8819](https://user-images.githubusercontent.com/45026713/63003516-77389400-beb7-11e9-90f9-de4f5a4619fe.JPG)

To the right of the word 'Dismiss' please show the current time.

GianniCarlo:
Agreed, I’ll look into fixing this as soon as possible. A little reasoning behind this: having the now playing view shown as a modal was a last minute change, given that if it was put as a two pager app, the now playing view will bug itself and be sligthly off screen and wouldnt present any information

GianniCarlo:
solved by #370

Serendigity:
Thanks for the quick action :)

PR #370, opened 2019-08-14
[Watch App] show clock and fix animation on now-playing-view

GianniCarlo:
Re: #369 

Instead of presenting the now playing view, now we push it. Results:
IMG_6307


GianniCarlo:
I'll be releasing 3.4.1 as soon as possible with this fix

PR #371, opened 2019-08-14
Release 3.4.1

GianniCarlo:
This only has the improvement introduced in #370 

Release notes:

## Improvements

– Smooth transition to the Now-Playing-View in the Watch App (it also shows the clock now).


GianniCarlo:
And we're live!

PR #375, opened 2019-08-27
Release 3.5.0

GianniCarlo:
CarPlay support!

This may be the smallest feature release in all of BookPlayer's history, but due to time constraints, I think it's better to push smaller releases more often instead of holding them off to do a big release after months.


GianniCarlo:
and it's live! 🎉

PR #384, opened 2019-09-19
Import from Web

### LINKED ISSUE #383: Add books from any website
I am writing code to add a new feature for BookPlayer to allow a user to import books from the web. It works well and just cleaning things up.  

Hopefully this will be useful for those who want to browse [librivox](https://librivox.org/search?primary_key=0&search_category=author&search_page=1&search_form=get_results) or keep their own books in a web accessible library. 

I'll submit a pull request shortly.. but I thought I'd post here in case someone wants to comment.

openaudible:
Allow users to import books from web pages. 
Also allows users to paste in URL for an audio book. 
Handles basic/digest authentication, but some cookie-based authentication will probably fail.

Might be nice to save last URL to user preferences... see: 
    static var last_url = "" // the last url we browsed to. Maybe save to preferences?

Closes #383


GianniCarlo:
Hi @openaudible ! sorry for the delay, I finally found time to test the feature, and I've got some comments:
1.- Regarding the browser screen, we need to have a space between the textfield and the top limit. We also need to change the '<' button and show an 'x' instead
Screen Shot 2019-09-29 at 5 38 48 PM
2.- The browser screen also needs the controls '<' '>' like what `SFSafariviewcontroller` provides (I found myself unable to go back to get another book from the previous list)
Screen Shot 2019-09-29 at 5 46 33 PM
3.- When I dismissed the controller, the import had already begun, and it didn't present the alert to create a playlist based on the downloaded files. Maybe we could hold off the files in the /tmp folder, and upon dismissing or finishing things up in the browser, move the files from /tmp to /inbox
Screen Shot 2019-09-29 at 5 44 03 PM
4.- The downloading popup could be improved, maybe remove the 0 and 100 from both sides, and add '%' after the number that increases
Screen Shot 2019-09-29 at 5 42 58 PM

GianniCarlo:
another improvement would be to present a list of recommended sites (like librivox) when the browser is first presented. Having the blank screen won't be of much help for users that don't know the best places to get free books

openaudible:
OK. Will attempt to make those changes. Agree all make sense. I was thinking about a bookmark page including, librivox, but I don't know any others. I'm mostly interested in allowing users to be able to point to their own personal libraries. But didn't want to go crazy on features. 
Again, I'm a rank beginner at iOS development, so not sure if I'll get the changes 100%. thanks for the feedback! Hope we can get it in... the only feature I wanted from BookMobile.

GianniCarlo:
It is a very useful feature indeed @openaudible thanks again for this. Let's get the import thing sorted out then (point 3). The other points regarding the layout I'll ping @pichfl to see if he's available and see what could be the better solution and I can do those if you'd like 👍

openaudible:
Sounds good. Maybe you can figure out the best way to create a playlist automatically when a .zip file is added? I tried to avoid digging into the logic import logic.  

As a new XCode/Swift programmer, I got a little discouraged trying to add 3 simple buttons (back/forward/close0. 

But I did find there is a webview change that enables back/forward with just a swipe that could 
```
        self.webview.allowsBackForwardNavigationGestures = true
```

Anyway, expect something in 3-4 days.

openaudible:
OK.. Posted a new set of commits! I am happy with the results. 

Added swipe left/right for browser back/forward. 
Added a "last browsed" page that lets a user go back to the previous web page for books. 
Added forward, back and close buttons. Icons taken from open source [brave browser](https://github.com/brave/browser-ios) 
Cleaned up some of the ImportFromWebPage code. I removed the 0 100 numbers and just have "Downloaded... 50%" as the text. 
Updated the readme to mention the new feature. 

I did not try to implement a bookmark system.  there is room for improvement..

I did not attempt to create a playlist while importing a zip file. That seems like the best default behavior.  Or treat a folder as a playlist. I'm not sure!

pichfl:
I'm not sure about this, as it became practically obsolete by the changes to Safari in iOS 13, which include a download manager that can actually save files which in turn can be moved into BookPlayer using existing interaction patterns. 

Since we'll support iOS 12 for another year, I think it's a good thing to have. Your implementation works as a good proof of concept. 

However, I would go a bit of a different direction, which would not involve us implementing a custom browse (less work & things to be broken), while still keeping the download mechanism in place (the core of this feature):

Instead of opening a browser, I'd just inquire for a URL to be downloaded, which could even be pre-filled from the clipboard. The download can be shown in the very same cell that we currently use to process a newly added file. If we detect basic-auth we can open a second dialog which asks for credentials. 

If we detect a file that doesn't not match our support file formats (which btw are a bit broader, any audio file could do) we should display a warning and abort.

If the file is an archive we should ask the same question we ask when an archive arrives via AirDrop, invoking the very same dialog should be a possibility in this scenario.

openaudible:
Hi Florian,

Appreciate your review and the work you've done on BookPlayer.

I am looking at the iOS 13 download manager. I wasn't able to figure out how to download MP3's. I hold click on an MP3 URL and I don't see an option to download. Clicking once starts playing the mp3, without any (obvious?) way to download the file. If I'm mistaken and its easy to save MP3's using iOS 13, I don't think we need a paste URL or browser.

The reason I like the web browser is I can save my book library location with the embedded user:pass in the http url. When I update my library, I can browse my books, click on the mp3 I want, and its in BookPlayer in seconds. Full disclosure, I am the author of a openaudible, a book manager that saves your audio book library to a web page.. So this PR makes both our products better, in my opinion. :) 

But understand my PR adds a bunch of complexity.  I would also have preferred moving the new code, storyboard, and assets into a new module that could be separate and deleted easily without changing the main storyboard. But this is my first attempt at Swift, which is why the code may appear awkward or awful. But swift is a really nice language.. Happy to get a chance to learn it. 

Anyway... open to suggestions.. but think the "paste url" is not a great solution. Copying, pasting, typing passwords, just seems like too many steps. But let me know! Happy to help if I can.

pichfl:
Re iOS 13 downloads: I only tested quickly on LibriVox, but doing a long press on the "play" icon that links to the mp3 file contains a download action as the second option in the long press menu. Same goes for the share sheet when viewing an mp3 file, which includes an option to "Save to files". 

----

Adding a generic web browser marks our app as adult content, which is something I'm not very happy about either.

I still think having a mechanism to download and import a file directly from a URL is a very good idea and it brings very interesting possibilities for integration with other environments as well. If we have a way to download a file from a regular URL, we can use that for other features as well: 

- allow importing a URL from an extended custom URL scheme. Eg. `bookplayer://import?user=foo&password=bar&url=http%3A%2F%2Farchive…` (Note: I used separate optional query params since `user:password@domain` format for web URLs has been deprecated in RFC 3986) 
- create a share sheet extension which allows direct download into BookPlayer (without going through the Files.app)

This way, any website or app can link directly into BookPlayer and send over critical information, including the website created by openaudible, which would indeed be a nice thing to have.

openaudible:
Oh- you are right, I forgot I was using Chrome and not Safari!  Doh!

There is room for improving the experience: 
If you click on the "Add Book" you could look in the downloads directory for new files and prompt the user.. "You have [3] new downloaded audio files. Move them to your library now? [Y/N]" This should be pretty safe as it looks like the download directory gets cleared frequently. (Zip files should probably be avoided.) And then the book would be imported then deleted from the download directory. Some basic checks for file exists. If I have some time, and you think it's worthwhile, I'll experiment.

Personally, I don't think a audiobook:// protocol handler would be worth the effort. I hate exposed passwords so I wouldn't want to implement or use. The user:pass@ was just a hack to not have to implement a real password strategy. 

A file extension handler is the correct way to implement it: (long click on MP3/m4b/zip and Open with... BookPlayer).  

OK, I'm fine closing and deleting this PR request!

pichfl:
- AFAIK there is no way to watch the downloads folder as it is outside of BookPlayers sandbox, but maybe I'm not quite up to date on that information.
- BookPlayer does have a custom protocol to start the last played book, so we would just extend on what's already there.
- You could do without exposing passwords by also allowing for other means of authentication like a one-time token generated per file and download. Which would be the most secure option, given that basic auth is sent in plain text as well. I mostly suggested the protocol as means for easy cross application communication which comes in handy for the scenario you had in mind for openaudible. Having a button there that triggers the protocol would work from any browser on the device. 
- Opening files is already implemented via the share sheet (and is one of the most basic features). This doesn't cover URLs, which is why we would need the download handling inside BookPlayer.

Just to make this clear: In an ideal world, all of the discussion above should have taken place in the issue you created and before any implementation. I don't enjoy raining on your parade and turning this down, but once more, I feel like there is a very useful feature in there.

If you enjoy the challenge, @GianniCarlo and would be ever so grateful if you would want take another jab at this.

openaudible:
Hi Florian,

I don't mind the rain. You made the right call. I should have discussed it more.. but I'm not sure we could have anticipated iOS 13 changing Apple's 12 year policy of not allowing MP3 downloads. 

My goal was to have BookPlayer be able to import a book from a secure web site without copy/paste and typing passwords. (BookMobile didn't handle passwords or https) So issue is solved. 

Will keep my eye on possible improvements in the future! Again, thanks for your hard work on the best Audio book player for iOS.

PR #408, opened 2019-12-27
Change AVAudioPlayer for AVPlayer

GianniCarlo:
This PR solves the not-so-random playback continuing after taking off the earpiece. It also lays the groundwork for streaming a book from a server with AVPlayer.

I also believe that this is a potential solution to those who have had the playback reset back to zero when playing the last book when the app is not in memory.

Feel free to comment and provide feedback 👌


GianniCarlo:
when changing the speed of the player, the audio sound strange, like with a double voice depending on the speed. Also AVPlayer works differently and setting the speed to anything != 0, it assumes it should start playing, I'll add a validation for that

GianniCarlo:
alright, both issues have been addressed. I think functionality wise, it's ready, I'll squash some commits and force-push this before doing the merge

PR #419, opened 2020-01-18
List view search

clint015:
Here is the initial commit for adding search to list views. It's a fairly basic implementation. Also included is:
- Bumped the minimum supported iOS version to 11.4
- Removed function that was deprecated in iOS 11.

Before merging, I'd like to smooth out a couple of rough edges, such as searching on a list in editing mode messes with the Select All/Deselect All button a bit. I just wanted to make sure I was going in the right direction first.

Looking forward to your feedback. Thanks!


GianniCarlo:
Hi @clint015 ! throughout this week I'll be able to test it and give you more feedback 👍. Thanks again for your contribution 🙂

BookPlayer/Library/PlaylistViewController.swift
```diff
@@ -13,7 +13,22 @@ class PlaylistViewController: ItemListViewController {
     var playlist: Playlist!
 
     override var items: [LibraryItem] {
-        return self.playlist.books?.array as? [LibraryItem] ?? []
+        let items: [LibraryItem] = self.playlist.books?.array as? [LibraryItem] ?? []
+
+        if let searchText = searchController.searchBar.text,
```
GianniCarlo: we could write this part with a `guard` statement and another minor improvement I think:

```swift
let items: [LibraryItem] = self.playlist.books?.array as? [LibraryItem] ?? []

guard let searchText = searchController.searchBar.text,
  !searchText.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
  return items
}

return items.filter { item in
  if let book = item as? Book {
    return book.title.contains(searchText) || book.author.contains(searchText)
  }

  // since all LibraryItem have title, we can skip the check of playlists
  // unless we plan on searching the string among the playlist contents, which could make sense 🤔
  return item.title.contains(searchText)
}
```

GianniCarlo: oh wait, this block only applies to playlists, and since we don't support nested playlists, the filter block could just be `item.title.contains(searchText)`

BookPlayer/Library/ItemListViewController.swift
```diff
@@ -24,16 +24,37 @@ class ItemListViewController: UIViewController, ItemList, ItemListAlerts, ItemLi
...
-    // TableView's datasource
     var items: [LibraryItem] {
         guard self.library != nil else {
             return []
         }
 
-        return self.library.items?.array as? [LibraryItem] ?? []
+        let items: [LibraryItem] = self.library.items?.array as? [LibraryItem] ?? []
+
+        if let searchText = self.searchController.searchBar.text,
```
GianniCarlo: Similar thing as with the first recommendation in the `PlaylistViewController` implementation

GianniCarlo:
sorry for my late review @clint015 , visually, I like how it looks 👌. In regards to functionality, I think we should:
- Filter ignoring the case of the string
- When filtering from the library, consider filtering the inner books from each playlist to see if there's a match

GianniCarlo:
There's another thing, that I can't seem to find why it happens. When I scroll (slowly) to hide the searchbar, it sort of jumps and triggers the transition from navigation big title to small title

AFBCFC84-C324-4CE9-8593-30BAB09FDF5B

Another thing is that when the navigation title is big, and the search bar is hidden, if you enter a playlist, then when popping back, the search bar is shown 🤔

clint015:
Yeah, I’m seeing that, too. I’ll take a look and see if I can figure out what is going on.

clint015:
I haven't been able to figure this out. I did find https://stackoverflow.com/questions/47649375/ios-11-large-navigation-bar-title-unexpected-velocity where people suggested setting `extendedLayoutIncludesOpaqueBars` to `true` fixes the issue, but I'm not able to determine if that has any affect. I'm user a slower MacBook than before (long story) and I'm not even sure I can consistently reproduce the issue.

GianniCarlo:
I'm closing this PR in favor of #647

PR #424, opened 2020-02-07
Release 3.6.2

GianniCarlo:
Release notes:

Improvements

- New speed options (3x, 0.9x, 0.5x)

Bugfixes

– Fixed being unable to pick a completed book from inside a playlist (it skipped to the next one)
– Fixed sleep timer (end-of-chapter option), it wasn't stopping in playlists when the current book finished
– Tapping mini artwork wasn't starting playback in the case where the book was already loaded


GianniCarlo:
It's live in the App Store 🎉

Issue #286, opened 2018-11-24
Localization


pichfl:
We should move the app to proper localization.

Our app statistics seem to suggest we have quite a few Russian and German users (and I'm not all of them). Also having a Spanish version won't hurt for the US market.

I can provide and maintain translation strings for German and I'm quite confident we'll find someone who wants to help us out with the Russion version.

Nibelungc:
Hi. I can help with Russian localization

pichfl:
@Nibelungc wohoo. That was even faster than I expected. Thanks a lot! We'll keep you posted about when and how.

GianniCarlo:
@Nibelungc awesome!

on another point @pichfl , some time ago I stumbled on this framework https://lokalise.co which is free for open source projects, let me know what you think about it (it might be overkill for just one app though)

ps. I'll do Spanish

pichfl:
@GianniCarlo I haven't tried them but having a managed solution for the translation keys makes for a much better translator experience and might help us a lot. Totally in favor of lokalise.co here.

GianniCarlo:
An update on this, we got a project going on https://lokalise.co and we've added all the keys that we're currently using in BookPlayer. @Nibelungc please let us know with which email you'd like to be invited to the localization project so you can contribute with Russian translations

GianniCarlo:
so far we've got covered:
- [X] English (Base)
- [X] German (@pichfl)
- [X] Spanish (@GianniCarlo)
- [X] Russian (@Nibelungc)

If you'd like to see the app localized in your language and contribute, please let us know so we can add you to the project

Nibelungc:
nibelungc@gmail.com

GianniCarlo:
@Nibelungc invitation sent, thanks for your help!

carcade:
Can help with russian localization.

GianniCarlo:
We'll be picking this up again to finally close and merge it for the next version (3.4.0) 🙂. We'll have more news on this throughout this month

devil418:
I can help with the polish version. I'm miki123211 on Google's email service. I'm a blind user, so not sure how my screen reader is going to handle the translation SAAS, though.

pichfl:
@devil418 we can always export a plain text translation file for you. We want to get some other features out of the way before we start this one, but we'll get back to you very soon!

GianniCarlo:
[...]

PR #427, opened 2020-02-12
Localization

GianniCarlo:
Good news!

This PR has the localization we've gathered so far in the platform https://lokalise.com . Since that point in time for #286 , the app has grown and we've added more features and more text that needs to be localized.

I'll be uploading in lokalise.com the new keys in the next few days, and after we get all the translations, then we'll be ready to finally deploy BookPlayer in 7 different languages :D

Feel free to run this branch and test the translations added so far


pichfl:
✅ German Translation.

It's a rough draft and I'll have to check for errors in the next days, but it's a start 🎉 so happy to see this.

PR #446, opened 2020-04-13
Sleep timer shortcuts support

GianniCarlo:
This includes proper support for the Shortcuts app for the Sleep Timer feature

IMG_CBC121337DB9-1

IMG_A11912CC0D9A-1


GianniCarlo:
This also raises our deployment target to iOS 12

PR #460, opened 2020-05-06
Sort by most recent

huynhxloi:
Added ability to sort by most recent.


GianniCarlo:
The project doesn't compile on my computer :/. I'm seeing some changes made to the `BookPlayer.xcodeproj/project.pbxproj` project file, which seems to be the reason TravisCI also is giving an error. It looks all related to the dependencies, I would suggest the following:
1.- Discard the changes made to the file `project.pbxproj`
2.- Run in your terminal the command `carthage bootstrap --platform ios` to setup the dependencies

Let me know if I can be of further help 👍

huynhxloi:
I did a fetch then checkout the `project.pbxproj` from the masters and ran that carthage command but there's still a lot of buildtime errors.

image

GianniCarlo:
We coud schedule a call to help you setup the environment 🤔. But yeah, it should work with carthage, also, there's a `;` semicolon missing in the new keys in the localized strings

huynhxloi:
Missing semicolon fixed.

A call would be very helpful. Do you have Microsoft Teams? If not, Zoom?

GianniCarlo:
we can do it via zoom @huynhxloi 👌, currently weekends (mornings and afternoons) work for me, my timezone is GMT-5

GianniCarlo:
closing in favor of #475

PR #473, opened 2020-07-03
ISSUE-387 : Dismissing the player vc if there during background import

mmurali88:
Proposed changes

If the player is open and files are imported via AirDrop or sharing from an external app, the import progress isn't shown. Dismissing the player / presenting the Library would change that.

Types of changes

- [x]  Bugfix (non-breaking change which fixes an issue)

Checklist

- [x]  I have read the CONTRIBUTING doc


GianniCarlo:
This works great 👌 Thanks @mmurali88

PR #475, opened 2020-07-04
ISSUE-459 : Adding Most recent sort method

mmurali88:
Proposed changes

Implement most recently played item sort method for both the library view and the playlist view. 

Types of changes

- [x]  Feature Implementation

Checklist

- [x]  I have read the CONTRIBUTING doc
- [x] Linter check and compiled locally before pushing 
- [x] Tested various sorting scenarios across both library view and the playlist view

Note : 

1. The Date type is required for sorting otherwise the sort is just happening by date and not by the time played.
2. Localized string is updated only for English and need help for other languages.

### DIFF
--- a/Shared/BookSortService+SortError+PlayListSortOrder.swift
+++ b/Shared/BookSortService+SortError+PlayListSortOrder.swift
@@ -4,21 +4,30 @@ final class BookSortService {
     public class func sort(_ books: NSOrderedSet, by type: PlayListSortOrder) -> NSOrderedSet {
         switch type {
         case .metadataTitle, .fileName:
-            return self.sort(books, by: type.rawValue)
+            let sortDescriptor = getSortDescriptor(by: type)
+            return self.sort(books, by: type.rawValue, sortDescriptors: [sortDescriptor])
+        case .mostRecent:
+            let sortDescriptor = getSortDescriptor(by: type, ascending: false)
+            return self.sort(books, by: type.rawValue, sortDescriptors: [sortDescriptor])
         }
     }
-
-    private class func sort(_ books: NSOrderedSet, by key: String, ascending: Bool = true) -> NSOrderedSet {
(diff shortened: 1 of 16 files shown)


Shared/BookSortService+SortError+PlayListSortOrder.swift
```diff
@@ -31,6 +40,7 @@ enum SortError: Error {
 public enum PlayListSortOrder: String {
     case metadataTitle = "title"
     case fileName = "originalFileName"
+    case sortByMostRecent = "lastPlayDate"
```
GianniCarlo, line 43: let's change `sortByMostRecent` to just `mostRecent`, so when calling the function `sort(by:type:...` we can have consistency with the other cases in this enum

GianniCarlo:
This works great 👌, Re translation, we do have an alternative 👀, we could use the already translated key `carplay_recent_title` which is just `Recent`, so we would need to rename that key to something more generic like `recent_title` and reuse it everywhere. Or we could go with `Most Recent`, in which case I would ask that the english translation be included in the rest of translated `.strings` files, so I can later update them with Lokalise.

I'm open to both options regarding the translations, so let me know 👍

Shared/BookSortService+SortError+PlayListSortOrder.swift
mmurali88: updated based on comments.

mmurali88:
> This works great 👌, Re translation, we do have an alternative 👀, we could use the already translated key `carplay_recent_title` which is just `Recent`, so we would need to rename that key to something more generic like `recent_title` and reuse it everywhere. Or we could go with `Most Recent`, in which case I would ask that the english translation be included in the rest of translated `.strings` files, so I can later update them with Lokalise.
> 
> I'm open to both options regarding the translations, so let me know 👍

I went with the 2nd option mainly because I couldn't test the carplay option. While we do the UI re-vamp I will make sure to update it .

For now I have added the "sort_most_recent_button" item to all localization files for you to update later.

GianniCarlo:
👌👌👌

PR #476, opened 2020-07-05
ISSUE-242 : Enabling chapters button only for multi chapter book

### LINKED ISSUE #242: Books with a single chapter should be handled as books without chapters
Easy enough to detect and fixing shouldn't be that hard after #231

mmurali88:
Proposed changes

Books with a single chapter should be handled as books without chapters. 

Types of changes

- [x]  Feature modification (non-breaking)

Checklist

- [x] I have read the CONTRIBUTING doc
- [x]  Linter check and compiled locally before pushing 
- [x] Tested multi chapter / single chapter / mp3 audio books
- [x] Unit tests passed.

Closes #242

### DIFF
--- a/BookPlayer/Player/PlayerViewController.swift
+++ b/BookPlayer/Player/PlayerViewController.swift
@@ -177,7 +177,11 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
             items.append(self.sleepLabel)
         }
 
-        self.chaptersButton.isEnabled = self.currentBook.hasChapters
+        guard let chapterCount = self.currentBook.chapters?.count else {
+            return
+        }
+        self.chaptersButton.isEnabled = self.currentBook.hasChapters && chapterCount > 1
+
         items.append(spacer)
         items.append(self.chaptersButton)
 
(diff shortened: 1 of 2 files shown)


BookPlayer/Player/PlayerViewController.swift
```diff
@@ -177,7 +177,13 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
             items.append(self.sleepLabel)
         }
 
-        self.chaptersButton.isEnabled = self.currentBook.hasChapters
+        if self.currentBook.hasChapters {
```
GianniCarlo, line 180: With this change, books that don't have chapters, have the chapters' button enabled, which shouldn't happen

mmurali88: Ah my bad. Tested with one chapter book before pushing but missed item without any chapter. I have updated the button's default state to disabled in the story board so this works fine. LMK if this works.

GianniCarlo: There's still one case not covered :/. If the user has autoplay in his library, or if it has in a playlist, a book with chapters, which is followed by a book with no chapters, when the player screen is reloaded for the next book (the one without chapters), the chapters button stays enabled

mmurali88: Updated the crash on the corner case you mentioned. Tested the following scenarios

Autoplay order :
1. Book with chapters > 1  -> No chapter - > 1 chapter (Library/playlist) -> 2 Chapter
2. No chapter - >Book with chapters > 1 ->  1 chapter (Library/playlist) -> 2 Chapter 
3. Library No chapter item -> PLaylist(Book with chapters > 1) -> Library(1 chapter item)  -> Library(2 Chapter)

All of these have been tested with and without autoplay. Works as expected.

BookPlayer/Player/PlayerViewController.swift
```diff
@@ -177,7 +177,11 @@ class PlayerViewController: UIViewController, UIGestureRecognizerDelegate {
             items.append(self.sleepLabel)
         }
 
-        self.chaptersButton.isEnabled = self.currentBook.hasChapters
+        guard let chapterCount = self.currentBook.chapters?.count else {
```
GianniCarlo, line 180: This `guard` here if the `count` ever resolves to `nil` then the button `avRoutePickerBarButtonItem` after this block would never be added. Testing it, I found out that for books w/o chapters, it resolves to `0`, but I think that it's better if we change this guard and make it more explicit by replacing it for:
```swift
let chapterCount = self.currentBook.chapters?.count ?? 0
```

Issue #299, opened 2019-01-07
Save interaction history as automatic bookmarks


cuddawudda:
I have been a Bookmobile user for a couple of years and there are a few features I would like to see added in Bookplayer. 

This request is for a history of 'bookmarks'. This has occasionally been invaluable when I have left the app playing or inadvertently skipped and I have been able to quickly go back to a previous position. 
What is the best method to share screenshots here?

In the meantime here's a link to a debit on the web. Note that majority are auto created, but some are manually created and text can be edited and colour coded. 

![576x768bb](https://user-images.githubusercontent.com/194641/50797381-e53f2c00-12d4-11e9-8134-56e29f08a37a.jpg)

Edit by @pichfl: inlined image

GianniCarlo:
I need this. Auto-created bookmarks should a huuge help, specially if one falls asleep. I'll revise the [roadmap](https://github.com/TortugaPower/BookPlayer/projects/1) and see where we can add the bookmarks

pichfl:
That is a neat idea! Totally in favor of this.

cuddawudda:
I know i raised this...
But is this a duplicate of  #127 ?

GitHubThrowaway31231:
I've recently started using this app (fantastic work btw!), and I went searching on Google to see if this feature was available which led me to this page.

I just wanted to add that I think this feature would be incredibly valuable, even if it started off as a very simple, non-interactive page which is effectively just an audit of playback actions (e.g. list with entries like "At 10:32am Playback Started at Position 3:32 on Book XYZ").

My use case is that I often listen to books when falling asleep, and so the app will end up being way ahead of where I actually listened to. An audit page would allow me to more easily jump back to a relevant time.

Is this feature likely to be on the horizon at some point? If I had a Mac I'd be tempted to give it a shot myself!

GianniCarlo:
Hi @AeroflotBhutanHymning ! we do plan to bring this feature alongside bookmarks, but it working as an audit page may be out of scope, since there may not be a point in storing past the last 10(?) interactions, and it'll just continue replacing themselves. 

Although we could easily store all, and just present the latest 🤔, and the audit page is an interesting point, for statistics to make them available for the user. We are already storing time spent listening in a day (I explain a bit in #373 ), so it can be queried for a future widget (about how much time one has spent listening to audiobooks), but it doesn't have that depth to distinguish which titles have taken more time. That is something that I'd really like to expand on and be able to show the user a better usage statistic of the app

[...]

Issue #127, opened 2018-06-03
Support for bookmarks


nameisken:
I saw they are no bookmark function be supported. I am wondering, does this because no API(AVFoundation) to support it or any other reasons? in my experience, the bookmark is an essential feature in audiobook players, so does it have any possibility to implement it or any trouble with it?

pichfl:
Thank you for your input! Bookmarks are tricky to get right, but could be very useful for non fiction and reference. If you can help us specify the bookmarks feature in more detail, we would be grateful. 

- Could you elaborate on what you expect from a bookmark feature? 
- How do you expect bookmarks to work? (How do you expect BookPlayer to create, edit, delete bookmarks? Is there a standard for this type of meta data?)
- Is there another app with that feature which works as you expect it to do?

We have #10 on [our Roadmap](https://github.com/GianniCarlo/Audiobook-Player/projects/1) which while rather different seems at least to be related to your request. 

So much for converting your ticket into a feature request, now to answer your question: No, API is not the reason this wasn't implemented yet. BookPlayer is much more than just the rather small part of the actual playback and isn't limited to the API surface of AVFoundation. 

The focus of BookPlayer is to provide an easy to use player for audiobooks. For the upcoming version support for Playlists, faster load times, and a more intuitive and cleaner UI were determined to be the most pressing points to work on. 

As with all feature requests I'm closing this ticket to focus on the work that needs to be done for the current release. It will remain active on our Roadmap and can and should be discussed further.

nameisken:
Thank you pichfl 
I'm using Apple's iBooks and Amazon's Audible application to listen ".m4b" and ".aax" files, they are good but not perfect.

"iBooks" has essential functionality and convenient, snappy UI operations. unfortunately, it doesn't support bookmark feature. 
"Audible" has more functionality and supports bookmark feature, unfortunately, it only supports its own cloud "aax" files, offline "aax" and m4b files will have odd chapter titles and no bookmark functions.

Why I'm so fancy the bookmark feature?
people are used to listening somewhere many times in an audiobook. An audible book same as a textbook which needs marks or underlines to warm up people's memory. this is a big difference for entertainment with music.

Why do I suggest to do it?
Why would people need another audible player when they already have Amazon and apple's player and they are good? this is a question. Except for the essential functions those players should all have(not too much), we should do on the difference which apple and Amazon don't, and the bookmark is the one.

I'm not sure but I would like to do some if I can.
[...]

PR #626, opened 2021-09-06
[BKPLY-60] Bookmarks support

GianniCarlo:
## Purpose

As part of our player redesign, we can now finally bring bookmarks support. We are also adding two types of 'automatic' bookmarks for when playback begins, and when the user skips in time

reference:
* https://github.com/TortugaPower/BookPlayer/issues/127 (bookmarks)
* https://github.com/TortugaPower/BookPlayer/issues/299 (auto bookmarks)

## Linear tasks

[[BKPLY-60] Bookmarks support](https://linear.app/bookplayer/issue/BKPLY-60/bookmarks-support)

## Approach

* Refactor controllers and views with `Storyboarded ` and `NibLoadableView`
* Bookmarks flow work similar to how Audible does their clips

* For the Coordinator approach, we only created `MainCoordinator`, we have yet to refactor the whole app flow to only use coordinators

## Screenshots

https://user-images.githubusercontent.com/14112819/132262149-90201451-74e8-4f59-a3cf-f95b974c6154.mov


GianniCarlo:
Updated bookmark icon

IMG_CAC5B89AE2BE-1

PR #645, opened 2021-10-06
Release 4.2.0 (Bookmarks support)

GianniCarlo:
This version finally brings bookmark support. The entire app has been reworked, so to be on the safe side, I think we'll use the rest of October to do an open beta with TestFlight. I just sent the build for review, so it should be up on TestFlight by this Friday more than likely, I'll post the link in this PR once it's live

Draft release notes:
Improvements

– Bookmarks are finally here!
– Redesigned Player with easier to access playback controls
– You can now export or share multiple items at a time of your library via Airdrop or your preferred app that can handle audio files
– The content of zip files will be imported automatically without needing a second prompt from the user

The entire app has been reworked and optimized too, so if there's anything behaving oddly, please feel free to report it at support@bookplayer.app


GianniCarlo:
The beta is live, link to join: https://testflight.apple.com/join/FLv3zLyN

GianniCarlo:
Feedback received via email: 
- Autoplay not working correctly when CarPlay is active (investigate)

Issue #531, opened 2021-01-02
Search library


deeohlaa:
Is there a search option available or coming soon to allow users search for books by title in their library?

pichfl:
A quick search here in the project would have brought up #105 which discusses just that. There is also Pull Request #419 which tried to implement the search but was never completed.

We are currently busy reworking the file structure of the Library behind the scenes, but a big Library update is planned in the near future 🙂

PR #647, opened 2021-10-11
feature[library]: search through books

geekygrappler:
## Purpose

- Add basic search to the library

## Linear tasks

#531 

- Not sure if I'm updating the data in the correct way, using `updateSnapshot` but it works 😅 
- Authors seemed to be under `book.details` on my simulator. Not sure if this is correct or could change with different data?
- I hard coded the placeholder "Search by Title or Author". Not sure how the localization works in the app, show me an example and I'll update it.

## Screenshots

https://user-images.githubusercontent.com/3627301/136791281-107b3521-0b0e-405a-98f0-9394e14727a8.mp4


geekygrappler:
Yes you are right, the snapshot is just 13 items.

Doctor-love:
I have no opinions regarding the technical implementation, but I really like the feature - I have quite many books in my library and this kind of function would be nice

GianniCarlo:
Closing this PR due to inactivity, as an additional note, to avoid the problems with the animation of the search bar, we can just have a magnifying glass button in the nav bar, and have it push a new screen dedicated to search results

PR #661, opened 2021-11-01
Fix artwork aspect ratio

GianniCarlo:
## Purpose

Portrait artwork was getting distorted in the player view

## Approach

Added a black background and changed scaleToFill to aspectFit. The black background is to keep the artwork 'shaped' as a square and have the airplay icon be part of the artwork too

## Screenshots

IMG_AB594B4A2A45-1

### DIFF
--- a/BookPlayer/Player/Player Screen/Views/ArtworkControl.swift
+++ b/BookPlayer/Player/Player Screen/Views/ArtworkControl.swift
@@ -80,7 +80,7 @@ class ArtworkControl: UIView, UIGestureRecognizerDelegate {
 
     self.artworkImage.clipsToBounds = false
     // artwork now has the main info regarding the title and author
-    self.artworkImage.contentMode = .scaleToFill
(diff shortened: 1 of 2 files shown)


pichfl:
@GianniCarlo Won't adding black to the artwork stand out quite a bit for light artworks and themes? I would recommend using transparent and letting the airplay icon float for now

GianniCarlo:
not really sure about not having the black background @pichfl , I'm attaching a screenshots:

light theme with black background

IMG_3DBF6293B0D5-1

light theme with no black

IMG_E3A868ADE017-1

pichfl:
Hmm I see what you mean. It would look very odd for a light cover though. 

Would you consider entertaining my absurd relation with code based color arrangements and try the average cover of the artwork instead? The Apollo dev had a post about getting that one much faster than we do now: https://christianselig.com/2021/04/efficient-average-color/

PR #662, opened 2021-11-01
Release 4.2.1

GianniCarlo:
Draft release notes:

Hot-fix!
– Fix autoplay not working
– Fix crash when opening the app due to duplicated identifiers
– Fix portrait artwork aspect ratio in player screen
– Add support to skip tracks in player screen

Please check out what we introduced in the previous version as well:

Improvements

– Bookmarks are finally here!
– Redesigned Player with easier access to playback controls
– You can now export or share multiple items at a time of your library via Airdrop or your preferred app that can handle files
– The content of zip files will be imported automatically without needing a second prompt from the user

The entire app has been reworked and optimized too, so if there's anything behaving oddly, please feel free to report it at support@bookplayer.app


GianniCarlo:
The app has been submitted for review, hopefully it's out by wednesday

PR #750, opened 2022-01-29
Revised speed controls

GianniCarlo:
## Purpose

After receiving feedback from the community, we're changing the UIStepper for a UISlider (similar to @s00d original proposal), in addition we're adding 3 quick action buttons (we may make these configurable in a later version)

## Screenshots

https://user-images.githubusercontent.com/14112819/151643036-4b125caf-6376-4c1d-aab2-2f01d79c058c.MP4


s00d:
image
Hello. For fine tuning, I suggest doing something like this, thanks for your work.

PR #853, opened 2022-11-01
New controls UI for sorting and selecting

GianniCarlo:
## Purpose

Replacing the Edit button with a bar that has both the Sort and Select buttons, make room for the Search button in the navigation bar, and also it helps discoverability of these additional actions for users that never tap on the Edit button

## Screenshots

IMG_21628C62F5CF-1


GianniCarlo:
Localization will be handled on v5.0.0 branch when it's near the first beta version

Issue #867, opened 2023-01-14
Bug: IOS - Audiobook gets deleted when opened via Files.app (or imported a second time)


crieke:
## Environment

 - iOS 16.2
 - App Version: 4.7.2-1

## Description

I am not sure, if this issue fits into this topic as well. As I've noticed a file deletion issue on iOS as well.
If I click on an m4b via the Files app in the `BookPlayer/Processed` folder, it opens up the BookPlayer app and asks to import that file again (even though it is already imported). 
- You have 2 choices then: If you cancel the process, the audiobook gets deleted and the already imported one can't be played anymore.
- You click Import (again) and have a duplicate of the audiobook in the Book Player library.

It'd be great, if BookPlayer would not try to import an already imported audiobook again when it is selected from the Files.app within BookPlayers own folder. Instead it could launch BookPlayer and start Playback of the corresponding audiobook.

## Steps to reproduce the behavior
1. Have BookPlayer installed on your phone with some imported Audiobooks 
2. Open Apples Files.app and browse "On My Phone"`/BookPlayer/Processed`
3. Click on any audiobook file
4. Import Screen in BookPlayer will pop up
5. Either select "Cancel" (-> Audiobook will no longer be available in Bookplayer)...
6. ...or select "Import" (-> Audiobook will be in the Library twice)

GianniCarlo:
🤔 are you accessing it via Files due to the search functionality? or is there another reason you go through there? that aside, yeah it makes sense to detect that the incoming file is already from the Processed folder, so it doesn't prompt importing the book again

crieke:
Actually, I tried to find the audiobook file in the Files app to share it with another device by airdrop (and was not aware of the option to share it within BookPlayer when swiping the audiobook in the Library). That led to the situation that I clicked some m4b files in the files app by accident.
I was a little bit puzzled about the deleted file afterwards...
But I could imagine, it might happen in other situations as well, when people just browse or (as you mentioned) search for files and click on it...

GianniCarlo:
ooooh ok, makes sense 👍

lamnatos:
Just to second this, I've had the same thing happen to me. Tap book in Files.app, select Cancel, book is deleted from app.

GianniCarlo:
The fix is included in the newly released v5.0.0 🙌

PR #897, opened 2023-04-23
#867 - Prevent importing files that are already in the library

Jeepston:
## Bugfix

When user taps on the audio file in the Processed folder, and then Cancel, file is deleted as part of the clean-up process. If user taps on Import, library ends up with duplicate file.

## Related tasks

https://github.com/TortugaPower/BookPlayer/issues/867

## Approach

In `CommandParser.parse` function add another check if file is in the Processed folder. If it is, stop import process (silently ignore)

### DIFF
--- a/Shared/CommandParser.swift
+++ b/Shared/CommandParser.swift
@@ -43,6 +43,10 @@ public class CommandParser {
 
   public class func parse(_ url: URL) -> Action? {
     if url.isFileURL {
+      guard !url.absoluteString.contains(DataManager.getProcessedFolderURL().absoluteString) else {
+        return nil
+      }
+      
       return Action(command: .fileImport, parameters: [URLQueryItem(name: "url", value: url.path)])
     }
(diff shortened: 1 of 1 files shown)


GianniCarlo (changes requested):
Sorry about my delay, I've been working on `v5.0.0` branch ironing out issues, and we're now really close 😁

Regarding the PR, the issue is still happening, I think you need to see into the screen `ImportViewController` → `ImportViewModel`, both delete functions `deleteItem(_:)` and `discardImportOperation` rely on the `ImportManager`, so we can add the check in there (probably in the `removeFile` function should be enough)

GianniCarlo:
ah, I didn't know that when merging the branch `v5.0.0` it would automatically close the PR, feel free to open it again pointing to `develop`. As I said, we're pretty close, and we're going to start testing the entire v5.0.0 internally this week 😁

PR #899, opened 2023-05-03
#867 - Prevent importing files that are already in the library

Jeepston:
## Bugfix

When user taps on the audio file in the Processed folder, and then Cancel, file is deleted as part of the clean-up process. If user taps on Import, library ends up with duplicate file.

## Related tasks

https://github.com/TortugaPower/BookPlayer/issues/867

## Approach

In `CommandParser.parse` function add another check if file is in the Processed folder. If it is, stop import process (silently ignore). Additional check when creating import operation in `ImportManager` in order not to duplicate files; and when cleaning up after the cancelled operation in order not to delete files from Processed folder.

### DIFF
--- a/Shared/Extensions/URL+BookPlayer.swift
+++ b/Shared/Extensions/URL+BookPlayer.swift
@@ -13,12 +13,6 @@ public extension URL {
         return (try? resourceValues(forKeys: [.isDirectoryKey]))?.isDirectory == true
     }
     
-    var isInProcessedFolder: Bool {
-        let absoluteUrl = resolvingSymlinksInPath().absoluteString
-        let processedFolderUrl = DataManager.getProcessedFolderURL().absoluteString
-        return absoluteUrl.contains(processedFolderUrl)
-    }
-
   // Disable file protection for file and descendants if it's a directory
   func disableFileProtection() {
     try? (self as NSURL).setResourceValue(URLFileProtection.none, forKey: .fileProtectionKey)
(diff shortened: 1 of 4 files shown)


Shared/Extensions/URL+BookPlayer.swift
```diff
@@ -12,6 +12,12 @@ public extension URL {
     var isDirectoryFolder: Bool {
         return (try? resourceValues(forKeys: [.isDirectoryKey]))?.isDirectory == true
     }
+    
+    var isInProcessedFolder: Bool {
```
GianniCarlo, line 16: I can confirm that this is working, but can we move this to the `DataManager`? by including it in the extension, we're adding a relation to the DataManager from the native URL struct, I would rather have the `DataManager` be the one who owns this check, we could have a class func like this and use that:

```swift
public class func isURLInProcessedFolder(_ url: URL) -> Bool {
    let absoluteUrl = url.resolvingSymlinksInPath().absoluteString
    let processedFolderUrl = getProcessedFolderURL().absoluteString
    return absoluteUrl.contains(processedFolderUrl)
  }
```

GianniCarlo (changes requested):

Shared/Extensions/URL+BookPlayer.swift
Jeepston: @GianniCarlo I've moved the code from URL extension to DataManager.

GianniCarlo (approved):
Thanks again @Jeepston 😁

Issue #394, opened 2019-10-20
Name playlist after folder/archive name instead of first file


mikolysz:
*feature*
* What is the current behavior?
When importing from a folder/zip into a playlist, the default playlist name is the name of the first file, which doesn't make sense.
* What is the expected behavior?
The playlist name should be the folder name, or, the name of the zip file, without the extension.
* What is the motivation / use case for changing the behavior?
The first file is often called "chapter1.mp3", which doesn't tell us much, especially when we have more than one book with that convention. A name based on the archive/folder would make much more sense here.

pichfl:
Very good idea! 

Notes to anyone who wants to take a stab at this: It is likely that this might require some refactoring in the ImportOperation class.

PR #906, opened 2023-05-27
Suggest folder name same as imported zip archive

Jeepston:
## Purpose

When importing zip archive, and then selecting import into New Folder, app suggests folder name same as first file in the archive. This PR changes the behaviour and now suggested folder name is the same as imported archive file name (without extension).

## Related tasks

https://github.com/TortugaPower/BookPlayer/issues/394

## Approach

1. Capture `suggestedFolderName` in `ImportOperation`
2. Pass it to `ItemListViewModel` which shows finished import Alert with suggested folder name


GianniCarlo (approved):
Nice! this looks great 🙌

Issue #920, opened 2023-06-19
Bug: ...double / triple catalogue entries


a-oe:
## BookPlayer on iPhone 

 - OS Version: [iOS 16.5]
 - App Version: [5.0]

## After importing new audiobooks via computer / USB, some library entries are double ore even triple in list
No iCloud connection (iCloud disabled).

## Steps to reproduce the behavior
1. Copy files via USB from Desktop Computer
2. View list of audiobooks in app
3. Mostly the last entrie are double. Sort order makes no difference.

GianniCarlo:
This looks to be related to #915 and a failed migration from the store, I'm copy / pasting my findings:

I was able to reproduce this issue. After a failed migration, the CoreData stack ends up in bad place, and does not work reliably, for example when trying to display the next page of items, it's ignoring the offset of the already loaded items, and that's why it's duplicating items.

To force a failed migration, I created a new schema version (v10), and tested going from model v8 to v10 (model v9 was introduced in app version 5.0.0). It is coded in a way that it will apply each model/schema version at a time, and I found a bug, that after it finishes migrating to v9, I was trying to load the data store to perform custom actions and populate folder information, the problem is that the expected version of the store is v10, and it was just migrated to v9, that's where it fails. So in the AppStore version, one possible scenario is that users that are experiencing this bug, went from schema v7 to v9.

I'm testing for ways to recover from this state, probably go to the previous version and retry the migration, but in any case, what mckayfackrell posted in this ticket, it's a good solution

> Same issue for me. I was able to locate all the files if I searched in BookPlayer and in the files app on iPhone under the processed folder.
>
> I moved my book files to a new folder in the files app and deleted the BookPlayer app. I redownloaded the app and the only way to import that worked was by selecting the add files (plus button) in the BookPlayer app and manually selected the files.

a-oe:
Thanks! I'll post a "close with comment", after succesful update.

GianniCarlo:
@a-oe got a build on TestFlight live right now with a fix for this (link to join: https://testflight.apple.com/join/xLDneJXN), please let us know if it fixes your issue, just a heads-up, folders that may have previously disappeared due to the bug, may appear with blank details on first launch, but on subsequent app launches they'll be populated correctly

a-oe:
Quick test in TestFlight: 
2 audiobook files loaded via USB: Both are double in the list view.
Sorry, not fixed.
[...]

Issue #918, opened 2023-06-18
Latest Update Stopped App From Working!


appsteve:
## Environment

 - OS Version: [e.g. 13.5.1]
 - App Version: [e.g. 3.9.0-1]

## Description

The latest update has stopped the app from opening! It is showing following error when trying to open the app: “Multiple validation errors occurred.” I even deleted the app, restarted my iPhone, and installed it again, but the app still gives the same error and craps out. I may also have lost all my data! Please fix and fully test the app before releasing a new update! Thanks

## Steps to reproduce the behavior
1. Restarted iPhone
2. Reinstalled app
3. Opening app gives following error and app goes no further (craps out). I’ve been using the app daily for a few years now, so this is really sad to see. Error: “Multiple validation errors occurred.”

appsteve:
Forgot to add versions:

iOS: 16.5
App version: 5.0.0 (latest as of June 18, 2023)

GianniCarlo:
we're looking into it, thanks for the report

nauni77:
I have the same problem. Even reboot or reinstall didn't help.

iOS: 16.5
App version: 5.0.0 (latest as of June 18, 2023)

Just for information: The device is a apple iPhone 12 Pro.

Thanks for your help!

Oli

karb0f0s:
Also have this issue. Is there a way to gather log/trace to investigate?

GianniCarlo:
it looks like a CoreData issue when trying to perform migrations from the previous data model to the new one, I've been trying to replicate but with no luck

(screenshot of core data errors enum)
<img width="988" alt="Screenshot 2023-06-19 at 09 29 54" src="https://github.com/TortugaPower/BookPlayer/assets/14112819/d2f32eae-a8b3-43e7-943b-f205f4ce593c">

Unfortunately as it is not crashing, Sentry is not reporting this case, I'll add logs with the [OS logger](https://developer.apple.com/documentation/os/logger), so they are visible with the Console app on MacOS for connected devices to help debug this in future cases. In the meantime, I'm adding this to the list of unrecoverable states we have, so the CoreData stack is reloaded and the app is at least accessible again (progress will be lost :/)

nauni77:
Hello, but why should this caused by a data mirgration process? The same error happens if I delete the complete app and install it from AppStore again. In this case IMHO there should be no data to migrate ... I'm wrong?

GianniCarlo:
[...]

PR #922, opened 2023-06-19
Possible fix for app stuck in launch

GianniCarlo:
## Bugfix

Clean install of the app is stuck at launch

## Related tasks

#918 

## Approach

- Change load up sequence to go ahead and load up the library
- Recreate mapping model
- Add logs


GianniCarlo:
This should also resolve #920 and folders having 0 duration

PR #934, opened 2023-06-25
WIP: Rework CoreData interface to address random crashes

GianniCarlo:
## Purpose

- There's been random crashes from CoreData, after looking into it, it's because the main context from CoreData is being accessed from different threads, and this overtime can cause random crashes. 
- Our current approach was serializing the `NSManagedObject` into a separate in-memory struct, to avoid relying on directly the DB objects for displaying the screens, and when we need to update objects, we would fetch them from the DB by id and make the updates and store them. The problem is that CoreData's `NSManagedObjectContext` are restrained to the Queue they were created on, so the main context that there's only one, it's tied to the main thread

## Approach

- Rework all the interface with an `async / await` wrapper, so we can fetch the context on the correct queue, to either fetch or update from the DB
- Use `Task` where needed to make use of the new `async` functions

- There are functions that are already executed on the main thread, so in theory we don't need the `async` wrapper, these will be identified as we continue to test these changes
- These changes are too big to go into the next release without proper testing, so I'll keep these separate, while I look into the import issues


GianniCarlo:
I'm closing this, as it's definitely a better idea to just migrate to Realm, instead of investing more time into CoreData

Issue #642, opened 2021-10-03
In player, display remaining time instead of full duration by default


rakleed:
# 🗣 Context

In the player on the right, display the remaining time instead of the full duration by default. Since in most iOS applications for listening to music and podcasts, the remaining time is displayed on the right (for example, -0:52). And when switching to your app, consistency is lost, making the app harder to use.

# 💬 Narrative

In the player on the right, display the remaining time instead of the full duration by default. Since in most iOS applications for listening to music and podcasts, the remaining time is displayed on the right (for example, -0:52). And when switching to your app, consistency is lost, making the app harder to use.

# 📝 Notes

# 🎨 Design

<details>
<summary>Screenshots</summary>

![IMG_1823](https://user-images.githubusercontent.com/19418601/135746699-034ea099-98ab-4805-8033-b25bf362dfa3.jpeg)

![IMG_1824](https://user-images.githubusercontent.com/19418601/135747085-b70f6b01-e090-4f7b-8742-1683581d2e44.jpeg)

</details>

# ✅ Acceptance Criteria

**GIVEN**
**WHEN**
**THEN**

# 🚫 Out of Scope

GianniCarlo:
Hi there @rakleed , you can already do this, just tap that same label, and it will toggle between total and remaining time. Also if you're listening to a book that has chapter metadata, you can tap the label in the middle to toggle between book duration and chapter duration (affecting both current time and max time)

rakleed:
@GianniCarlo thanks! Can you display the remaining time by default? Since this is the default behavior for most other iOS apps, and this is what the user expects to see.

GianniCarlo:
Once you change it once, the app remembers your choice, and will use that for the next items you play 👍

rakleed:
I understand, but usually users don't change the default settings. Why don't you want to make the behavior I have proposed as the default option, taking into account the previously indicated arguments?

GianniCarlo:
I'm focused on other things regarding the upcoming version, if anybody else wants to contribute with a pull request with the code to add that change, they are welcomed to do it, I also personally prefer the remaining time option, but that is my preference

rakleed:
Can you then reopen the issue?

GianniCarlo:
This is now live on v5.0.3 in the AppStore 🎉 thanks again @Jeepston

PR #939, opened 2023-06-27
display remaining time by default

Jeepston:
## Purpose

Users are used to see remaining time on the progress bar of the player. This PR changes default behaviour of the app and sets UserDefaults value for `remainingTimeEnabled` to `true` by default, whereas before, it was `false`, and total book time was displayed.

## Related tasks

https://github.com/TortugaPower/BookPlayer/issues/642

## Approach

Set default value for `Constants.UserDefaults.remainingTimeEnabled` in `DataInitializerCoordinator.setupDefaultState` function


GianniCarlo (approved):
This is definitely something that has been asked via mail support as well, as it's not clear that the functionality is available, and that they could switch between total and remaining time from either settings or the player. Hopefully with this we'll get less mails about it 😅, thanks @Jeepston !

PR #943, opened 2023-06-28
simplify enum constants; remove unused code

Jeepston:
## Purpose

Simplifies usage of enum constants; removes unused code

## Approach

- convert all read-only enum constants from cases to static let variables, which allows to drop usage of `.rawValue` in all places where that constant is needed
- removed unused constant enum related to artwork theme colours, since everything theme-related was moved to `SimpleTheme`


GianniCarlo:
@Jeepston  It looks like the build is failing because the `PlayerSettingsViewModelTests.swift` and `PlayerManagerTests.swift` still use the `.rawValue`

Screenshot 2023-06-28 at 18 03 47

Screenshot 2023-06-28 at 18 02 28

Jeepston:
Ah yes, I knew I've missed something :D For some reason, project didn't have Test scheme generated. After adding and compiling for tests, I was able to fix it.

GianniCarlo (approved):

Issue #796, opened 2022-05-25
Manage your files / sorting


oldman63:
Under Settings / Manage your files

The file list is not sorting!

Please Add sorting options or sort the list by file name

Thx

GianniCarlo:
Currently it's sorted by file size, but we could add a sort button at the top right corner to toggle between size and by file name 🤔

oldman63:
Ok, thx

GianniCarlo:
v5.0.4 is now live on the AppStore 🙌

PR #946, opened 2023-06-30
#796: Add a button to sort files on "Manage your files" screen

Jeepston:
## Purpose

Requested enhancement to add sort option in Storage Management screen. Current sort options are Size (default), and Title. Further options can be added (full path, creation date etc.)

## Related tasks

https://github.com/TortugaPower/BookPlayer/issues/796

## Approach

Initially, rewrite entire screen to SwiftUI. Correspondently, update ViewModel, Coordinator and Tests. Delete StorageViewController, StorageTableViewCell and unused StoryBoard Scenes.

Then, add a sort button similar to that in Library, which shows Picker with sort options: Size (default), and Title. After selecting new option, ViewModel re-sorts files and stores the selection in UserDefaults.

Translation for one of the sort option ("Size") is missing. There is a "TODO" comment in `StorageView` about it

## Screenshots

Simulator Screenshot - iPhone 14 Pro - 2023-06-30 at 17 43 25


BookPlayer/Settings/Storage/StorageRowView.swift
```diff
@@ -0,0 +1,90 @@
...
+      .padding(15)
+
+      VStack(alignment: .leading, spacing: 2) {
+        Text(item.title)
+          .font(Font(Fonts.title))
+          .multilineTextAlignment(.leading)
+          .foregroundColor(themeViewModel.primaryColor)
+
+        Text(item.path)
+          .font(.subheadline)
```
GianniCarlo, line 39: I think this `.subheadline` and the next one, needs to be a size smaller, to emulate what we have on the library cells, that the title is bigger, and the rest of labels a bit smaller

GianniCarlo (commented):
There's an issue with the callbacks from the rows, I'm not getting the alert shown when tapping them (see video)

https://github.com/TortugaPower/BookPlayer/assets/14112819/ce160e50-dbf1-42e3-b215-1a7a52cbe2b0

Other than that and the details label sizes I mentioned in my other comment, this is a great step forward in the SwiftUI migration process 🥳 ! also, good call on making the delete and fix icons a bit bigger, they're much easier to tap now

Jeepston:
Interesting 🤔 It worked initially. I’ll try to fix it tomorrow

Jeepston:
@GianniCarlo I've pushed the fix. In which I've learned that SwiftUI has some troubles handling multiple `.alert` modifiers which present `Alert` struct. Newer alert modifiers do not have this issue, but they are only available from iOS 15. So I've rewritten alert handling - there is only one modifier in the View, but it shows different Alerts depending on user action.

Finding number 2: SwiftUI defines some data type `LibraryItem` somewhere, so there are conflicts when using app's Core Data class `LibraryItem` in SwiftUI. It should be pre-pended with `BookPlayerKit.` 🙄

Finding number 3: when deleting files from Storage Management, it can leave "empty" LibraryItem in the DB, which shows in the Library, but cannot be played. I think this should be fixed, and I will submit it in the separate PR

GianniCarlo:
* Yeah it's sort of frustrating that modifiers are gated to newer iOS releases :/
* I ran into something similar with the `Section` enum, I'll keep this in mind when I continue with the Realm migration, so I will probably add a prefix `BP` to the data models, thanks for the heads up
* This one is by design, I think it was one of the requests I got, that they wanted to offload the files, and keep the progress, so they could potentially in the future import that book again, and it will remap to the object in the library

I'll take a look at the new changes in the morning, I got held up today by trying to figure out why Realm was not compiling, and it was because of the build setting `APPLICATION_EXTENSION_API_ONLY` set to `true`, I reported the issue on their repo, so hopefully there's a workaround to not disable that flag

GianniCarlo (approved):

PR #991, opened 2023-08-24
Fix Recent Books Widget's spacing error and text alignment.

thai-d-v:
## Bugfix

The widget distributes inequality space for book items when fewer than 4 books and 1 has a long title.
image

The widget should display equally like this.
image

More than that, I suggest changing the title alignment to make multiline titles look more balanced.
image


GianniCarlo (approved):
Thanks @thai-d-v for the PR!

PR #994, opened 2023-08-25
Fix indentation for all Swift files.

thai-d-v:
## Bugfix

I re-indent all Swift files (2 space width).


BookPlayer/AppDelegate.swift
```diff
@@ -23,12 +23,12 @@ import WatchConnectivity
 class AppDelegate: UIResponder, UIApplicationDelegate {
   static weak var shared: AppDelegate?
   var pendingURLActions = [Action]()
-
```
GianniCarlo, line 26: Is there a way to leave unmodified the empty lines without any whitespace?

thai-d-v: I use the “Re-Indent” command in Xcode. Could you show me how you format your code? I will try to re-format it using your tools and settings. 😁

harrisi: There are a few options, but the setting for this is in Settings (`⌘,`) > Text Editing > Editing > While Editing, and setting both the "Automatically trim trailing whitespace" and "Including whitespace-only lines".

Sometimes this doesn't work very reliably with re-indent, so you may need to find and replace using something like sed with something like `s/^\s+$//g`.

thai-d-v: @GianniCarlo I fixed it. Thank you @harrisi.

GianniCarlo (approved):
Nice!

PR #1001, opened 2023-09-08
I refactor `Coordinator`,  `BaseViewController`, `BaseViewModel` and some other classes to make themself and their subclasses smaller and safer.

thai-d-v:
## Purpose

- Make class interfaces smaller and more precise.
- Generic code is more static and safer than inherent code.
- Make code more expressive with more clear abstraction.

## Approach

- I move class members to protocols.
- Make classes conform and inherit only abstractions they need.

- Only 7 coordinators need `AlertPresenter`
- Only 1 coordinator needs `UINavigationControllerDelegate`
- `BaseViewController`, `BaseTableViewController`, and `BaseViewModel` are completely replaced by new protocols.
- I keep some old code with deprecated messages to make the situation clear.


BookPlayer/Utils/BaseViewController.swift
```diff
@@ -17,14 +18,45 @@ class BaseViewController<T: Coordinator, U: BaseViewModel<T>>: UIViewController
...
+}
+
+extension ViewControllerProtocol {
+  func accessibilityPerformEscape() -> Bool {
+    self.viewModel.dismiss()
+    return true
+  }
+}
+
+@available(*, deprecated, message: "Use `TableViewControllerProtocol` instead.")
```
GianniCarlo, line 34: I think we there are some improvements we can do based on this file:
- Rename `ViewControllerProtocol` to `MVVMControllerProtocol`
- Delete `BaseTableViewController`
- Delete `TableViewControllerProtocol`, and use instead the renamed `MVVMControllerProtocol` (the missing `accessibilityPerformEscape` was an oversight on my part)

BookPlayer/Coordinators/LibraryListCoordinator.swift
```diff
@@ -190,12 +190,23 @@ class LibraryListCoordinator: ItemListCoordinator {
     }
   }
 
-  override func interactiveDidFinish(vc: UIViewController) {
+  func navigationController(_ navigationController: UINavigationController, didShow viewController: UIViewController, animated: Bool) {
```
GianniCarlo, line 193: Some context here: the idea is to move away from the detach / attach pattern of the coordinators, and have them be kept in memory by the view models, so when the flow from the coordinator is dismissed, it's automatically deallocated, and we would no longer need to have these checks

GianniCarlo (commented):
I like the direction of this PR, as it helps kickstart some of the cleaning up that I've been meaning to do

I left a comment of some improvements we can do before merging the PR, and another about some context (no action needed there)

For additional context, my initial goal when I introduced the `BaseViewController` was to force every new controller to have a view model and a coordinator in charge of presenting it. I later realized that not every controller needed a coordinator 😅, as a coordinator should handle flows, not individual screens (e.g. ButtonFreeCoordinator shouldn't exist, and the PlayerCoordinator should handle routing to the ButtonFreeViewController). So my goal was to start transitioning the view models, to no longer have a reference to the coordinator, and instead have a `onTransition` callback, where they can notify the route to whoever instantiated and is listening (one example of modifying to the new intended structure is `SearchListViewModel`)

BookPlayer/Utils/BaseViewController.swift
thai-d-v: I just pushed the new code, I hope it's ok. 😁

GianniCarlo (approved):
Nice! thanks @thai-d-v

PR #1023, opened 2023-10-09
Add Dead Man's Switch settings.

cvakiitho:
## Purpose

Take this as an enhanced feature request 🙂, 

I use BookPlayer almost exclusively before sleep - I always use timer so I don't have to seek and find where I left - since it's always just last 30 minutes or so. But sometimes, I don't fall asleep before the timer expires. Then playback stops of course, so I hit play on my airpods - but then - the second playback doesn't have the timer. 
I would have to reach for my phone, or it will play till my airpods lose power... I know, first world problems :) 

This is really my first time using xcode/swift/whatever, so I hope I didn't break anything, but it works fine on my phone, and in ios Sim.

Not sure about the name, but I found this one ok.

- Write a few sentences here outlining the purpose of the PR
- **Use good markdown styling**, including `inline code` highlighting for variable or function names, or code blocks for any multiline snippets of code.
- If you are not familiar with Github flavored markdown, you can find it [here](https://guides.github.com/features/mastering-markdown/)

## Approach

- Tell the reviewers the general approach that was taken
- Include any new classes, services, endpoints, etc. that were added.

- Anything to watch out for in reviewing this PR, or anything that requires special attention? Put it in here.
- If there are particular things to review, make a bullet list here

## Screenshots

- Would a screenshot(s) help communicate the improvements, issue being fixed? Then include a screenshot.
Screenshot 2023-10-09 at 21 43 11

### DIFF
--- a/BookPlayer/Player/Player Screen/PlayerViewModel.swift
+++ b/BookPlayer/Player/Player Screen/PlayerViewModel.swift
@@ -394,6 +394,7 @@ class PlayerViewModel: ViewModelProtocol {
         }
       )
     )
+    
 
     let formatter = DateComponentsFormatter()
 
@@ -430,7 +431,26 @@ class PlayerViewModel: ViewModelProtocol {
         }
       )
     )
-
+    
+    if(SleepTimer.shared.getSticky()){
+      actions.append(
+        BPActionItem(
+          title: "sleeptimer_option_sticky_on".localized,
+          handler: {
+            SleepTimer.shared.setSticky(stickyState:false)
+          }
+        )
+      )
+    } else {
+      actions.append(
+        BPActionItem(
+          title: "sleeptimer_option_sticky_off".localized,
+          handler: {
+            SleepTimer.shared.setSticky(stickyState:true)
+          }
+        )
+      )
+    }
     actions.append(BPActionItem.cancelAction)
 
     sendEvent(.sleepTimerAlert(
--- a/BookPlayer/Player/PlayerManager.swift
+++ b/BookPlayer/Player/PlayerManager.swift
@@ -739,6 +739,8 @@ extension PlayerManager {
 
       handleSmartRewind(currentItem)
 
+      handleStickyTimer()
+      
       fadeTimer?.invalidate()
       shakeMotionService.stopMotionUpdates()
       boostVolume = UserDefaults.standard.bool(forKey: Constants.UserDefaults.boostVolumeEnabled)
@@ -770,6 +772,12 @@ extension PlayerManager {
       let newPlayerTime = max(CMTimeGetSeconds(self.audioPlayer.currentTime()) - rewindTime, 0)
 
       self.audioPlayer.seek(to: CMTime(seconds: newPlayerTime, preferredTimescale: CMTimeScale(NSEC_PER_SEC)))
+    }
+  }
+  
+  func handleStickyTimer() {
+    if(SleepTimer.shared.getSticky()){
+      SleepTimer.shared.restartTimer()
     }
   }
 
--- a/BookPlayer/Player/SleepTimer.swift
+++ b/BookPlayer/Player/SleepTimer.swift
@@ -30,7 +30,7 @@ final class SleepTimer {
   /// Current time left on the timer
   @Published public var state: SleepTimerState = .off
   /// Last manually set sleep timer
-  private var lastActivedState: SleepTimerState = .off
+  private var lastActiveState: SleepTimerState = .off
   /// Default available options
   public let intervals: [TimeInterval] = [
     300.0,
@@ -39,6 +39,8 @@ final class SleepTimer {
     1800.0,
     3600.0
   ]
+  
+  public var sticky = false
 
   /// Publisher when the countdown timer reaches the defined threshold
   public var countDownThresholdPublisher = PassthroughSubject<Bool, Never>()
@@ -91,11 +93,19 @@ final class SleepTimer {
 
   // MARK: Public methods
 
+  public func setSticky(stickyState: Bool){
+      sticky = stickyState
+  }
+  
+  public func getSticky() -> Bool{
+    return sticky
+  }
+  
   public func setTimer(_ newState: SleepTimerState) {
     /// Always cancel any ongoing timer
     reset()
     state = newState
-    lastActivedState = newState
+    lastActiveState = newState
 
     switch newState {
     case .off:
@@ -114,9 +124,9 @@ final class SleepTimer {
       NotificationCenter.default.addObserver(self, selector: #selector(self.end), name: .chapterChange, object: nil)
       NotificationCenter.default.addObserver(self, selector: #selector(self.end), name: .bookEnd, object: nil)
     }
-  }
+    }
 
   public func restartTimer() {
-    setTimer(lastActivedState)
+    setTimer(lastActiveState)
   }
 }
--- a/BookPlayer.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/BookPlayer.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/devicekit/DeviceKit.git",
       "state" : {
-        "revision" : "66837ecf1516e41fd4251bbb684dc4b1997f08ab",
-        "version" : "5.1.0"
+        "revision" : "e1c258c6070ddd6465265fe4692f29717404aede",
+        "version" : "5.2.1"
       }
     },
     {
@@ -23,8 +23,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/iosdevzone/IDZSwiftCommonCrypto.git",
       "state" : {
-        "revision" : "d824371e670bd57eb456bbc41139b4997f7207b8",
-        "version" : "0.13.1"
+        "revision" : "47f9747d12137cd59d783ab376e3ccab7206319b",
+        "version" : "0.16.1"
       }
     },
     {
@@ -32,8 +32,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/onevcat/Kingfisher.git",
       "state" : {
-        "revision" : "c75584ac759cbb16b204d0a7de3ebf53ea6b304d",
-        "version" : "7.9.0"
+        "revision" : "3ec0ab0bca4feb56e8b33e289c9496e89059dd08",
+        "version" : "7.10.2"
       }
     },
     {
@@ -41,8 +41,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/cbpowell/MarqueeLabel.git",
       "state" : {
-        "revision" : "bc00d4cbff7f6c416035e3d16c21885452cd159e",
-        "version" : "4.3.1"
+        "revision" : "ae3cf7c647dd7f67f1946658285f5f0ce2476caf",
+        "version" : "4.4.0"
       }
     },
     {
@@ -50,8 +50,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/RevenueCat/purchases-ios.git",
       "state" : {
-        "revision" : "ec48897eff464c040bf774741b5f1d5a97053436",
-        "version" : "4.25.1"
+        "revision" : "ff4907a2b01fb6e0e891f36e9a14421a5500cc04",
+        "version" : "4.31.9"
       }
     },
     {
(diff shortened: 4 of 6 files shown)


GianniCarlo:
@cvakiitho I've been thinking about this, for context, what I had planned before, was to have a 'sticky' sleep timer, so anytime you resume playback, the last sleep timer would be set. This means reworking the UI for the sleep timer in the player, so you could toggle on/off this feature, and also toggle on/off the 'shake' to restart the sleep timer (and playback) which is currently on by default. 

I bring this up because your use case is not covered by a sticky timer really, but rather allowing the restart of the timer by restarting playback shortly thereafter (shake threshold is of 1 min I think). So I think the easier path on this task would be along those lines, so we avoid adding a different setting for the same goal which is to restart playback after sleep timer ends

If you're up for it, what I'm thinking about is
- Create another service to handle restarting playback after sleep timer
- Replace `ShakeMotionService` from `PlayerManager` with the new service, and move shake service into the new service
- Move timeout timer logic from `ShakeMotionService` to the new service
- Where `.bindShakeObserver` is being called, we would start the timer threshold on the new service
  - Starting the timer threshold would internally start the shake motion service listener for shake events
- On `play()` we would start the sleep timer if the threshold is active (or in the future, when the sticky timer feature is enabled too), and do cleanup on the shake motion service to stop events just in case

let me know what you think

cvakiitho:
I was thinking along those lines alsi, but "sticky timer" is exactly what this dead man's switch is right? 

So UX wise you'd want this option to exist in the "Timer modal" inside player, rather then in settings? 
UX wise as a user I don't really care I think, but having it in player Timer modal is probably slightly better. 

As for 

> If you're up for it, what I'm thinking about is
    Create another service to handle restarting playback after sleep timer
    Replace ShakeMotionService from PlayerManager with the new service, and move shake service into the new service
    Move timeout timer logic from ShakeMotionService to the new service
    Where .bindShakeObserver is being called, we would start the timer threshold on the new service
        Starting the timer threshold would internally start the shake motion service listener for shake events
    On play() we would start the sleep timer if the threshold is active (or in the future, when the sticky timer feature is enabled too), and do cleanup on the shake motion service to stop events just in case

I'm not really sure I'm swift skilled enough for this, but I can take a look - probably next week, as I'm on a vacation Thursday - Monday

cvakiitho:
As for shake to restart, how that feature even works? I never made sense of it, I've tried to shake my phone in several settings and it did nothing? is it like - shake after playback stops due to timer?

GianniCarlo:
> I was thinking along those lines also, but "sticky timer" is exactly what this dead man's switch is right?

Yes, it refers to the same thing

> So UX wise you'd want this option to exist in the "Timer modal" inside player, rather then in settings?
UX wise as a user I don't really care I think, but having it in player Timer modal is probably slightly better.

Right, that would be more desirable, as it enables space for the shake-related toggle as well. I brought the sticky timer thing up and the shake feature, because I think that your use case of 'resuming playback with the same sleep timer, after the current sleep timer ends and stop playback', actually falls into the same goal as the shake to resume feature, the only difference, is that instead of shaking, you're just resuming playback from your airpods, and we should include that as part of the intent of the user to resume playback with same sleep timer

> As for shake to restart, how that feature even works? I never made sense of it, I've tried to shake my phone in several settings and it did nothing? is it like - shake after playback stops due to timer?

Regarding how it works, once the timer ends, you'll hear the sound begin to fade away, from that point on to aprox 1 min you can shake the device to restart playback and sleep timer, the time window could probably be less depending on iOS and how much time it'll keep the app working in background with no sound playing. If you have selected the option of end of chapter, the volume won't fade, but it will pause at the end of chapter, and you can still shake your phone (no need to unlock it)

cvakiitho:
Reopening with implementation of `Sticky` option in timer modal.

cvakiitho:
Screenshot 2024-01-18 at 20 43 15

The downside of having it here, is, there is no visual hint that it's on right now, maybe some icon could be added🤔?

BookPlayer/Player/SleepTimer.swift
```diff
@@ -17,6 +17,7 @@ enum SleepTimerState: Equatable {
   case off
   case countdown(TimeInterval)
   case endOfChapter
+  case sticky
```
GianniCarlo, line 20: I don't think this one goes here, as the states are meant as _running_ states, so they are mutually exclusive, and a user can have the sticky preference toggled to On, while having the sleep timer running

BookPlayer/Player/SleepTimer.swift
```diff
@@ -39,6 +40,10 @@ final class SleepTimer {
     1800.0,
     3600.0
   ]
+  
+  public var lastInteral: TimeInterval = 300
```
GianniCarlo, line 44: we don't need this property as we already have the `lastActivedState` (just noticed my typo on that var naming 😅), that one is being set in `setTimer(_:)`, but it's being set always, I think we can add a conditional so we don't overwrite it if the user turns off the timer manually

BookPlayer/Player/PlayerManager.swift
```diff
@@ -772,6 +774,19 @@ extension PlayerManager {
...
+    let interval = SleepTimer.shared.lastInteral
+    if(SleepTimer.shared.sticky == true){
+    
+      if SleepTimer.shared.state == .off {
+        SleepTimer.shared.setTimer(.countdown(interval))
+      } else {
+        SleepTimer.shared.restartTimer()
+      }
+    }
+  }
```
GianniCarlo, line 789: we would only need to check if the sticky preference is On, and in that case call `restartTimer`, and it should automatically set the last timer that was active

BookPlayer/Player/SleepTimer.swift
```diff
@@ -39,6 +40,10 @@ final class SleepTimer {
...
+  
+  public var sticky = false
```
GianniCarlo, line 46: We'll need a setter for updating the stored value when the user taps on the alert sheet, and a getter to check on the stored value when resuming playback

BookPlayer/Player/SleepTimer.swift
```diff
@@ -113,6 +119,12 @@ final class SleepTimer {
       donateTimerIntent(with: .endChapter)
       NotificationCenter.default.addObserver(self, selector: #selector(self.end), name: .chapterChange, object: nil)
       NotificationCenter.default.addObserver(self, selector: #selector(self.end), name: .bookEnd, object: nil)
+    case .sticky:
+      if (sticky){
+        sticky = false
+      } else {
+        sticky = true
+      }
```
GianniCarlo, line 127: This would no longer apply

BookPlayer/Services/ActionParserService.swift
```diff
@@ -149,6 +149,8 @@ class ActionParserService {
       SleepTimer.shared.setTimer(.off)
     case -2:
       SleepTimer.shared.setTimer(.endOfChapter)
+    case -3:
+      SleepTimer.shared.setTimer(.sticky)
```
GianniCarlo, line 153: Re parsing a command for sticky, we can handle that in a future PR, to possibly setup an action to update user preferences

BookPlayer/Player/Player Screen/PlayerViewModel.swift
```diff
@@ -430,7 +435,15 @@ class PlayerViewModel: ViewModelProtocol {
         }
       )
     )
-
+    
+    actions.append(
+      BPActionItem(
+        title: "sleeptimer_option_sticky".localized,
```
GianniCarlo, line 441: while not the ideal UI for a preference, we can make it work if we have a separate string to denote the preference is On, and another when the preference is Off

BookPlayer/Player/Player Screen/PlayerViewModel.swift
```diff
@@ -430,7 +435,15 @@ class PlayerViewModel: ViewModelProtocol {
...
+        handler: {
+          SleepTimer.shared.setTimer(.sticky)
```
GianniCarlo, line 443: This would instead call the setter for the sticky preference mentioned in another comment

BookPlayer/Player/PlayerManager.swift
```diff
@@ -772,6 +774,19 @@ extension PlayerManager {
       self.audioPlayer.seek(to: CMTime(seconds: newPlayerTime, preferredTimescale: CMTimeScale(NSEC_PER_SEC)))
     }
   }
+  
+  func handleSticky(_ item: PlayableItem) {
```
GianniCarlo, line 778:
```suggestion
  func handleStickyTimer() {
```
No need to pass in the `PlayableItem`

GianniCarlo (commented):
Thanks for coming back to this @cvakiitho , I left some comments in the PR, I think with just some small tweaks we'll be ready 👍

The highlight is about treating the sticky flag as a preference and not as a sleep timer state, but let me know if I should add more details to any of my comments

BookPlayer/Player/SleepTimer.swift
cvakiitho: damn `lastInteral` typo 😅

cvakiitho: yeah that's true, I probably abused the sticky state that caused me to use another `last*` variable to actually get the last interval...

BookPlayer/Player/Player Screen/PlayerViewModel.swift
cvakiitho: right yeah, I thought that might be better - will try to implement that.

BookPlayer/Player/PlayerManager.swift
cvakiitho: Hmm it seems to me like that doesn't really work, seems to not restart the timer when getSticky return true 🤔

```
func handleStickyTimer() {
    
    if(SleepTimer.shared.getSticky()){
      SleepTimer.shared.restartTimer()
    }
  }
  ```

Comment (GianniCarlo) [reply BookPlayer/Player/PlayerManager.swift]: hmmmm 🤔 I guess between app launches it wouldn't work, but it should work after setting the first timer? (let me know if it still doesn't work) if that's the case we could store the last active timer so on init we set that value on `lastActivedState` 🤔

Comment (cvakiitho) [reply BookPlayer/Player/PlayerManager.swift]: ah ok, it seems like playback is actually crashing in my simulator for whatever reason. So I suppose I can't test if it works, but I think it should.

**Comment by cvakiitho** [comment]
I added `✓` to the sticky option when it's on, I suppose it could be Sticky when on, and Not Sticky when off? Not sure, 

-My simulator is throwing on play button so I can't really te-
ok that was with bad sample mp3, I actually tested it now, and it works fairly well. 

Just need to thing of a czech translation :D

File: BookPlayer.xcodeproj/project.pbxproj [hunk 10]
```diff
@@ -3842,9 +3840,9 @@
 				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_IDENTITY = "Apple Development";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_STYLE = Automatic;
```
GianniCarlo, line 3843: this file

BookPlayer/Intents.intentdefinition
```diff
@@ -95,15 +95,25 @@
 					<key>INEnumValueName</key>
 					<string>oneHour</string>
 				</dict>
+                                <dict>
```
GianniCarlo, line 98: this file

BookPlayerShareExtension/Info.plist
```diff
@@ -16,10 +16,10 @@
         $extensionItem.attachments,
         $attachment,
         (
-        ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.audio&quot;
-        || ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.folder&quot;
-        || ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.movie&quot;
-        || ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;com.pkware.zip-archive&quot;
+        ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.audio"
```
GianniCarlo, line 19: this file

Shared/CommandParser.swift
```diff
@@ -209,6 +209,8 @@ public class TimeParser {
       return 3600
     case .endChapter:
       return -2
+    case .sticky:
```
GianniCarlo, line 212: this file

GianniCarlo (commented):
Overall LGTM, the only thing remaining is if you can remove the committed changes for the files I left a comment for? just so I can merge the PR

GianniCarlo:
> Just need to thing of a czech translation 🫠

I can take over the translations haha, my workflow for that is to add them to lokalise.com and fill them with Google translate (it has a button to autofill the missing values), and when the contributors ping me that they updated a language, I replace the values and add them to the next release

cvakiitho:
Awesome, thanks a lot!

GianniCarlo (approved):

PR #1050, opened 2023-11-30
[BookPlayerShareExtension] Fix app group entitlement

moeadham:
Currently hard coded to tortugapower. Updated to use BP_BUNDLE_IDENTIFIER

## Bugfix

- Fix ShareExtension app group entitlement from hard coded value


GianniCarlo (approved):
Thanks for catching this @moeadham !

Issue #1057, opened 2023-12-17
Bug: Localisation of shortcuts not working


SimonHaas:
## Environment

 - OS Version: 17.2
 - App Version: 5.1.0-23

## Description

In the Shortcuts-app the localisation does not work despite the correct strings being present in the source code. I only tested with german.

Translation for "intent_lastbook_play_title"
https://github.com/TortugaPower/BookPlayer/blob/develop/BookPlayer/de.lproj/Localizable.strings#L300

The last string is displayed correctly.

![BAFF7B12-DE9E-442C-BF78-89AB0605A2FC_1_102_o](https://github.com/TortugaPower/BookPlayer/assets/22013217/1c1052e8-0f88-4352-9e3c-adff1eed8c88)

GianniCarlo:
Thanks for the catch @SimonHaas , I only checked the pre-built actions that AppIntents provide in the shortcuts app if you search the app, and thought everything was good, I'll see if there's anything extra to be done to get the translations working when adding actions to a shortcut

![IMG_442222C50D7E-1](https://github.com/TortugaPower/BookPlayer/assets/14112819/3e5d3991-6917-440c-9038-86d8d6c4ab6c)

GianniCarlo:
The fix is now live in the beta (link to join: https://testflight.apple.com/join/xLDneJXN)

GianniCarlo:
This is now live on the AppStore

PR #1059, opened 2023-12-18
Fix translations displayed in Shortcuts app for list of actions

GianniCarlo:
## Bugfix

- The app intents were not being translated inside the Shortcuts app when showing the list of available actions

## Related tasks

#1057 

## Approach

- Remove the table parameter from the title's initialization

- I'm including a change related to the previous PR, it's just grouping the conditionals under a single function

## Screenshots

IMG_9A104298B5D5-1
IMG_BB00940A610B-1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,8 @@ jobs:
     - name: Swiftlint
       run: swiftlint
     - name: Set Xcode version
-      run: sudo xcode-select -s "/Applications/Xcode_15.0.1.app/Contents/Developer"
+      run: sudo xcode-select -s "/Applications/Xcode_15.1.app/Contents/Developer"
     - name: Resolve dependencies
       run: xcodebuild -resolvePackageDependencies
     - name: Build and Run tests
-      run: xcodebuild -scheme BookPlayer test -testPlan Unit\ Tests -destination 'platform=iOS Simulator,name=iPhone 15,OS=17.0.1'
-
+      run: xcodebuild -scheme BookPlayer test -testPlan Unit\ Tests -destination 'platform=iOS Simulator,name=iPhone 15,OS=17.2'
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -22,5 +22,5 @@ jobs:
     - name: Resolve dependencies
       run: xcodebuild -resolvePackageDependencies
     - name: Build and Run tests
-      run: xcodebuild -scheme BookPlayer test -testPlan Unit\ Tests -destination 'platform=iOS Simulator,name=iPhone 15,OS=17.0.1'
+      run: xcodebuild -scheme BookPlayer test -testPlan Unit\ Tests -destination 'platform=iOS Simulator,name=iPhone 15,OS=17.0'
```
SimonHaas, line 25:
```suggestion
      run: xcodebuild -scheme BookPlayer test -testPlan Unit\ Tests -destination 'platform=iOS Simulator,name=iPhone 15,OS=17.2'
```
https://github.com/TortugaPower/BookPlayer/actions/runs/7262525331/job/19785852622#step:7:48

This change breaks the build. there is no OS=17.0

GianniCarlo: yeah I have to circle back to this, in the original build the tests never kicked off for some reason [[job]](https://github.com/TortugaPower/BookPlayer/actions/runs/7249398846/job/19767546775), so I was testing with [macos13-arm64](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-arm64-Readme.md) but the job never started for that, so I'm guessing it's only for some paid plan. I'll test out with 17.2 and hopefully the tests start without problems

GianniCarlo: yup that did it, thanks @SimonHaas

SimonHaas:
@GianniCarlo Thanks for your quick action!

SimonHaas:
@GianniCarlo As seen in your first screenshot, some strings are still in english. Maybe they are not available / translated to german yet? I will look into it myself the next days.

GianniCarlo:
@SimonHaas ah yes, those are the old shortcuts from SiriKit (they don't have a translation, I think I never figured out the proper way to translate them as they were defined in a separate special type of file for Siri intents),but they are being replaced by AppIntents, and once we drop support for iOS 14 and iOS 15, we can safely delete them from the project

Issue #1077, opened 2024-01-18
Add using URL


EddyG50:
## Environment iPhone 13

 - OS Version: 17.1.2
 - App Version: [5.1.1-28

## Description

I’m trying to us Add in BookPlayer then Download From URL. I have an mb4 file posted in Dropbox. When I paste the URL in BookPlayer it recognizes the file, looks like it’s downloading, but the fil shows 0mb in library and does not play. I also tried to zip the .m4b file but no luck with that.

How do I post a .m4b file that can be downloaded with URL directly into BookPlayer?

This is a great app! Thanks!

## Steps to reproduce the behavior
1. Post .m4b in DropBox & copy URL
2. Go to Add > Download From URL in BookPlayer 
3. Paste URL
4. Add to Library

GianniCarlo:
Hi @EddyG50 ! You have to make sure that the download URL is the actual book, many services like Google Drive, actually share first a link that shows the 'representation' of the file on the browser with the 'real' download button with the url in that page. If you can share the link, I can test it out also and see what I can find

EddyG50:
Thanks Gianni,
 
Below are links to both the .m4b and .zip - in both DropBox and Google Drive:
I'm using a small test file I created in Audiobook Builder App. I've also tried with a larger file.
I'm happy to archive it anywhere it can downloaded into BookPlayer from.
 
 
DropBox:
- m4b - [ https://www.dropbox.com/scl/fi/o3j73jtsqdby4tfzrnjqr/Test.m4b?rlkey=v8uk00ifbg61996ypo178jr1m&dl=0 ]( https://www.dropbox.com/scl/fi/o3j73jtsqdby4tfzrnjqr/Test.m4b?rlkey=v8uk00ifbg61996ypo178jr1m&dl=0 )
 
- .zip - [ https://www.dropbox.com/scl/fi/2euk64x6xd1abanutpb3m/Test.m4b.zip?rlkey=5pterrend5amv9d10swk7220w&dl=0 ]( https://www.dropbox.com/scl/fi/2euk64x6xd1abanutpb3m/Test.m4b.zip?rlkey=5pterrend5amv9d10swk7220w&dl=0 )
 
 
Google Drive:
[...]

PR #1163, opened 2024-08-02
Fix download from URL for Google Drive and Dropbox

nithya2511:
## Bugfix

For URLs to download books in Google Drive and Dropbox, the URL provided by the user needs to be fixed up before a download can be attempted.

Tested book download from Google Drive and Dropbox URLs for individual files and zip files.

## Related tasks

Fixes: #1077

## Things to be aware of

The fix is for Google Drive and Dropbox only; OneDrive download URLs are non-trivial.

### DIFF
--- a/BookPlayer/Library/ItemList Screen/ItemListViewModel.swift
+++ b/BookPlayer/Library/ItemList Screen/ItemListViewModel.swift
@@ -925,11 +925,13 @@ class ItemListViewModel: ViewModelProtocol {
           BPActionItem(
             title: "download_title".localized,
             inputHandler: { [weak self] url in
-              if let bookUrl = URL(string: url) {
-                self?.handleDownload(bookUrl)
-              } else {
+              do {
+                if let bookUrl = try self?.getDownloadURL(for: url) {
+                  self?.handleDownload(bookUrl)
+                }
+              } catch {
                 self?.sendEvent(.showAlert(
-                  content: BPAlertContent.errorAlert(message: String.localizedStringWithFormat("invalid_url_title".localized, url))
+                  content: BPAlertContent.errorAlert(message: error.localizedDescription)
                 ))
               }
             }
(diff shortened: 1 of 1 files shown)


BookPlayer/Library/ItemList Screen/ItemListViewModel.swift
```diff
@@ -1311,6 +1311,43 @@ extension ItemListViewModel {
       )
     ))
   }
+
+  func getDownloadURL(for givenString : String) -> URL? {
```
GianniCarlo, line 1315: let's make this a throwable function, so we return a non-optional URL. I think we can cover the throwable errors with just a `invalidURL` one, which we already have the localized strings for. It would go something like:

```swift
guard 
  let givenURL = URL(string: givenString),
  let hostname = givenURL.host
else {
  throw SomeErrorEnum.invalidURL
}

switch hostname {
  case "drive.google.com":
    return try getGoogleDriveURL(for: givenURL)
  case "dropbox.com", "www.dropbox.com":
    return try getDropboxURL(for: givenURL)
  default:
    return givenURL
}
```

We could also fallback to the original URL for both cases if the try fails for either dropbox or google drive functions

BookPlayer/Library/ItemList Screen/ItemListViewModel.swift
```diff
@@ -1311,6 +1311,43 @@ extension ItemListViewModel {
...
+        return getGoogleDriveURL(for: givenUrl)
+      } else if hostname == "www.dropbox.com" {
+        return getDropboxURL(for: givenUrl)
+      }
+    }
+    return nil
+  }
+
+  func getGoogleDriveURL(for url: URL) -> URL?
+  {
```
GianniCarlo, line 1329: minor nit, can we keep the opening brace on the same line as the function declaration?

BookPlayer/Library/ItemList Screen/ItemListViewModel.swift
nithya2511: `getGoogleDriveURL()` does not have failure path -- if we are returning the original URL -- so didn't add a try there. But otherwise I believe I have incorporated your suggestion.

GianniCarlo (approved):
Thanks @nithya2511 for the PR 🙌, I'll include it in the next release

PR #1304, opened 2025-06-06
Player manager audio session setActive error 

Kevincamp:
## Bugfix

Failed to activate the audio session

### DIFF
--- a/BookPlayer/Player/PlayerManager.swift
+++ b/BookPlayer/Player/PlayerManager.swift
@@ -11,7 +11,6 @@ import BookPlayerKit
 import Combine
 import Foundation
 import MediaPlayer
-import Sentry
 
 // swiftlint:disable:next file_length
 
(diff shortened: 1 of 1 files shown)


BookPlayer/Player/PlayerManager.swift
```diff
@@ -727,61 +728,75 @@ extension PlayerManager {
...
+                do {
+                    let audioSession = AVAudioSession.sharedInstance()
+                    try audioSession.setCategory(
+                        AVAudioSession.Category.playback,
+                        mode: .spokenAudio,
+                        options: []
+                    )
+                    try audioSession.setActive(true)
+                } catch {
+                    SentrySDK.capture(error: error)
```
GianniCarlo: I think we should leave the fatalError, as I've reached the scenario where the app can't enable an audio session, and no amount of retrying could get me past it, only force quitting the app would enable it to activate the audio session (I can't reproduce it lately, but still, just to be safe)

BookPlayer/Player/PlayerManager.swift
```diff
@@ -727,61 +728,75 @@ extension PlayerManager {
...
+            
+            if !autoPlayed {
+                handleAutoTimer()
+            }
+            
+            fadeTimer?.invalidate()
+            shakeMotionService.stopMotionUpdates()
+            
+            // Move UserDefaults access to background
+            Task.detached {
```
GianniCarlo: any particular reason to move this to a background thread?

BookPlayer/Player/PlayerManager.swift
Kevincamp: got it.

PR #1335, opened 2025-09-01
Release v5.12.0

GianniCarlo:
Draft release notes:

Improvements

– Reworked the entire Library tab (we now use the latest tools provided by Apple)
– Improved design for the item details screen too

Bugfixes

– Player: fix the chapter context not turning off/on when updating the flag in the settings screen

If you'd like to contribute with translations to other languages, please reach us at support@bookplayer.app


BookPlayer/Coordinators/MainCoordinator.swift
```diff
@@ -83,111 +79,77 @@ class MainCoordinator: NSObject {
...
-
-    startSettingsCoordinator(with: tabBarController)
+    let vc = AppHostingViewController(
+      rootView: MainView {
+        self.showSecondOnboarding()
+      } showPlayer: {
+        self.showPlayer()
+      } showImport: {
+        self.showImport()
+      }
```
jeremygrenier, line 89: I feel using a view model that will handle those action will be cleaner that use closure here.

BookPlayer/Coordinators/MainCoordinator.swift
```diff
@@ -58,22 +70,6 @@ class MainCoordinator: NSObject {
   }
 
   func start() {
```
jeremygrenier, line 72: ❓ Any reason we delay the creation of the view controller instead of doing it directly part of the init?

BookPlayer/Library/ItemDetails/ItemDetailsView.swift
```diff
@@ -0,0 +1,144 @@
...
+
+  @Environment(\.hardcoverService) private var hardcoverService
+  @Environment(\.dismiss) private var dismiss
+  @EnvironmentObject private var theme: ThemeViewModel
+
+  init(initModel: @escaping () -> ItemDetailsViewModel) {
+    self._viewModel = .init(wrappedValue: initModel())
+  }
+
+  var body: some View {
```
jeremygrenier, line 33: I think this body could be split in sub property or sub component. It is pretty tough to read.

BookPlayer/Library/ItemDetails/ItemDetailsView.swift
```diff
@@ -0,0 +1,144 @@
...
+        ItemDetailsHardcoverSectionView(viewModel: viewModel)
+          .listRowBackground(theme.secondarySystemBackgroundColor)
+      }
+
+      Section {
+        EmptyView()
+      } footer: {
+        VStack(alignment: .leading) {
+          Text(viewModel.originalFileName)
+          Text("\(Int(viewModel.progress * 100))% " + "progress_title".localized.lowercased())
```
jeremygrenier: I think it would be better to have the % part of the `progress_title` as an arg.

Also you could use iOS 15 new format API `viewModel.progress.formatted(.percent)` directly.

BookPlayer/Library/ItemDetails/ItemDetailsView.swift
```diff
@@ -0,0 +1,144 @@
...
+      isPresented: $viewModel.showHardcoverRemovalAlert,
+      presenting: viewModel.hardcoverAlertPayload
+    ) { payload in
+      Button("hardcover_remove_keep_it", role: .cancel) {
+        Task {
+          await viewModel.assignNewSelection(payload.newSelection)
+          viewModel.handleSaveAction(loadingState) {
+            dismiss()
+          }
+        }
```
jeremygrenier, line 74: ❓ Why do you do anything if the user cancel?

BookPlayer/Library/ItemDetails/Views/ItemDetailsArtworkSectionView.swift
```diff
@@ -0,0 +1,77 @@
...
+
+  var body: some View {
+    Section {
+      HStack {
+        if let image = image {
+          Image(uiImage: image)
+            .resizable()
+            .cornerRadius(4)
+            .aspectRatio(contentMode: .fit)
+            .frame(width: 100, height: 100)
```
jeremygrenier, line 28: Should you use `@ScaledMetric` for the size? Probably `ScaledMetric(relativeTo: .title)` so the image doesn't scale like crazy with dynamic type.

BookPlayer/Library/ItemList/Models/ItemListSearchScope.swift
```diff
@@ -0,0 +1,26 @@
...
+  var title: LocalizedStringKey {
+    switch self {
+    case .all:
+      return "All"
+    case .books:
+      return "books_title"
+    case .folders:
+      return "folders_title"
+    }
+  }
```
jeremygrenier: You can drop the `return` 🙈

BookPlayer/Library/ItemList/Models/LibraryNode.swift
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  var folderRelativePath: String? {
+    switch self {
+    case .root, .book:
+      return nil
+    case .folder(_, let relativePath):
+      return relativePath
+    }
+  }
```
jeremygrenier: :ditto: drop the `return`

BookPlayer/Library/ItemList/Models/ListStateManager.swift
```diff
@@ -0,0 +1,45 @@
...
+  private(set) var globalToken: Int = 0
+  private var tokens: [Scope: Int] = [:]
+  private var payloads: [Scope: Int] = [:]
+
+  public var isSearching = false
+  public var isEditing = false
+
+  func reloadAll(padding: Int = 0) {
+    payloads[.all] = padding
+    globalToken &+= 1
```
jeremygrenier: ❓ Do you really need `&+=`? I cannot see a scenario where we will hit the Int overflow.

BookPlayer/Utils/Extensions/AppHostingViewController.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by Gianni Carlo on 13/8/25.
+//  Copyright © 2025 BookPlayer LLC. All rights reserved.
+//
+
+import BookPlayerKit
+import SwiftUI
+
+final class AppHostingViewController<Content: View>: UIHostingController<Content> {
+  override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
```
jeremygrenier, line 13: Do you need this if you already have it part of the `AppNavigationController`?

BookPlayer/Utils/Extensions/VerticalLabelStyle.swift
```diff
@@ -0,0 +1,24 @@
...
+//  VerticalLabelStyle.swift
+//  BookPlayer
+//
+//  Created by Gianni Carlo on 22/8/25.
+//  Copyright © 2025 BookPlayer LLC. All rights reserved.
+//
+
+import SwiftUI
+
+struct VerticalLabelStyle: LabelStyle {
```
jeremygrenier, line 11: I think it could be great to have preview at least for those reusable component.

BookPlayer/Utils/Extensions/View+BookPlayer.swift
```diff
@@ -36,3 +37,39 @@ extension View {
...
+        Spacer().frame(
+          height: playerState.loadedBookRelativePath != nil
+            ? spacing
+            : Spacing.M
+        )
+      }
+  }
+}
+
+extension View {
```
jeremygrenier, line 56: ❓ Can `MainView` directly handle the safe area adjustment when the mini player is visible?

BookPlayer/Utils/ThemeViewModel.swift
```diff
@@ -13,6 +13,8 @@ import Themeable
 
 class ThemeViewModel: ObservableObject, Themeable {
   @Published var theme: SimpleTheme
+  @Published var defaultArtwork: Image?
+  @Published var defaultUIArtwork: UIImage?
```
jeremygrenier: ❓ Any reason to store both? Can you just have `@Published var defaultArtwork: UIImage?` and just use `Image(uiImage: defaultArtwork)` when need be.

`theme.defaultUIArtwork` seems to only be used once so you can also do the opposite and use `ImageRenderer(content: defaultArtwork).uiImage` or try to get rid of it altogether. I feel in the meantime storing only the `UIImage` is probably best.

GianniCarlo: Nice! I wasn't aware of the `ImageRenderer`

BookPlayer/Utils/Extensions/AppHostingViewController.swift
GianniCarlo: yes, the `AppNavigationController` will eventually be deleted when we migrate the player screens to SwiftUI, and the only way to force the orientation lock is via this bridge with UIKit which hosts the `MainView`, the options available in SwiftUI were not good enough :/

BookPlayer/Library/ItemList/Models/ItemListSearchScope.swift
GianniCarlo: force of habit 😅 , I'll update this too

BookPlayer/Utils/Extensions/View+BookPlayer.swift
GianniCarlo: I tried, but it doesn't quite work with how the safe area inset is already being used for the mini player, I can use a ZStack to overlay it, but I would still need to add the padding inside each tab root view

BookPlayer/Library/ItemDetails/Views/ItemDetailsArtworkSectionView.swift
GianniCarlo: I'm not quite sure I follow this one, it wouldn't scale right now with dynamic type, only the buttons on the right? or do you mean that we should support scaling the artwork?

BookPlayer/Library/ItemDetails/ItemDetailsView.swift
GianniCarlo: I think it's because that Jellyfin alert short-circuits the update logic, so no matter the selection, there could still be a pending update to the general artwork

BookPlayer/Coordinators/MainCoordinator.swift
GianniCarlo: this just follows the pattern from the other coordinators, to wait for `start` to actually create and push/present the view

BookPlayer/Coordinators/MainCoordinator.swift
GianniCarlo: hmmm I though about this, and these will eventually be removed when the SwiftUI migration is complete, I'll leave it as an incentive to continue with the migration haha

PR #1363, opened 2025-09-28
Add missing German localizations.

schaitl:
I noticed some missing German localizations, so I just decided to give it a go :)

I also noticed that the German localization is not consistent as to if the user is addressed formally ("Sie") or informally ("Du). I'd like to correct this as well (I guess informal is fine as it's dominating the current localization) and would issue another PR for that.


GianniCarlo:
Thanks @schaitl ! I'll merge this in a bit 🥳

schaitl:
Happy to help, thanks for the quick merge 🤘

PR #1365, opened 2025-10-02
Monospaced time fonts for elapsed and remaining time in player view

nm-f:
## Bugfix
Having used the app for a bit I noticed that the elapsed/remaining time labels were varying in width, depending on the string displayed. Other audio applications solve this with monospaced fonts.

## Approach

Overriding the font to be monospace can successfully be done for the `currentTimeLable` within `viewDidLoad`, however that doesn't appear to work for the `maxTimeButton` as every call to `maxTimeButton.setTitle` resets the style.

Not sure if this is the correct solution, as iOS is something I have limited experience with.

## Screenshots

It's not immediately apparent on the screenshots, but notice how the colon chars line up as the fonts are now monospaced.

monospace


GianniCarlo (approved):
This looks good, thanks @nm-f ! I'll include it in the next release

Issue #1182, opened 2024-09-15
Smart rewind and chapters


rakheshster:
## Feature request

- It would be good if smart rewind was chapter aware. So it rewinds 30 seconds _or_ to the start of the chapter if that's shorter. 

## Narrative

- Smart rewind is very useful but say I am 15 seconds into a new chapter and stop, later when it resumes it starts from the last 15 seconds of the previous chapter. Would be good if instead it started from the beginning of the chapter I am on. Or at least if there was some toggle to set this as a preference. 

**GIVEN:**
- Provide the context of the application
**WHEN:**
- Describe the action or trigger
**THEN:** 
- Describe the resulting change of the state in the application

## Design

_Provide design example, screenshot, text input/output etc_

arthurbrunborg:
Yes please. I would love this!

GianniCarlo:
this is now live in the AppStore, thanks again @lyallcooper !

Issue #976, opened 2023-07-24
Variable Smart Rewind


Lego296:
## Feature request

I really enjoy the smart rewind option, but please make it more customizable. 

30 seconds is a lot, switch between 10/20/30 sec should work for most users

gmaddry:
I agree this would be useful. Even better for dynamic option would allow setting the time for rewind.

BartemX:
Maybe also add an option to always rewind 2 seconds after a pause, Voice does this on android and it's one feature I miss.

GianniCarlo:
this is now addressed, you can set the interval, and also the amount we rewind per time passed is different, so it should feel more natural, thanks again @lyallcooper !

Issue #717, opened 2021-12-19
Feature request: quick rewind


freedmand:
## Feature request

- A boolean player controls setting option called "Quick rewind" (or similar) that when enabled causes every pause action to rewind a short amount of seconds (5 or 10 seconds)

## Narrative

- Often users are listening to audiobooks in hands-free settings, where their only effective control is to pause or resume playback (often via a bluetooth headset). If for whatever reason you mishear a word or zone out while listening in a hands-free setting, it can be effectively impossible to trigger a rewind action. This setting would allow the user to quickly rewind by simply pausing and resuming playback.

    **GIVEN:**
- The "Quick rewind" setting is enabled in player control settings
**WHEN:**
- The user presses pause
**THEN:** 
- The book is rewound a few seconds

## Design

<img src="https://user-images.githubusercontent.com/306095/146682714-b26f3c22-7452-458c-b382-3fb7ae4bbd63.jpeg" alt="settingspage" width="200"/>

An additional toggle setting on this screen, potentially directly below the "Smart rewind" setting:

Title (en): "Quick Rewind"
Default: false
Description (en): "Rewind 5 seconds after each pause action"

nobe4:
An alternative to this might be to generalise the "smart rewind" feature and give it 2 settings:
- After `Xs` of pause
- Rewind `Ys`

In my personal case, the smart rewind doesn't help, I end up rewinding manually a few minutes each time I pick up. But an automated smaller rewind would be appreciated :)

anohren:
Is it common for Bluetooth headsets to not provide a way to jump back/rewind? I've only used Sony headphones, but there I triple click play/pause to issue a jump backwards, etc. with only a single button.

Issue #624, opened 2021-09-05
Customized Smart Rewind for forgetful listeners


checkoutmyspideysuit:
# 🗣 Context

Smart rewind feature is not long enough for people with memory issues

# 💬 Narrative

I love using the app to listen to audiobooks but as a person with memory issues the smart rewind feature (rewind 30 seconds after not playing for >10 minutes) does not rewind enough for me to remember what is going on in the story. Personally, I often have to rewind ~90 seconds back to figure out what is going on because I forgot already. 

# 📝 Notes

n/a

# 🎨 Design

custom time setting for smart rewind.

I

PR #1382, opened 2025-10-18
Improve smart rewind behavior

lyallcooper:
## Purpose

This PR contains contains a few fixes:

**Don't smart rewind past beginning of chapter**
Currently, smart rewind will always put you back 30 seconds, even if you are very close to the start of a chapter. This is probably not the behavior that most users would expect or desire. So instead we can limit rewinding so that we never cross backwards over a chapter boundary. Users have requested this fixed in #1182.

**Add smart rewind interval setting**
This adds a new setting that allows users to specify the maximum amount of time they want the smart rewind feature to skip backwards by. The setting is only shown when smart rewind is enabled. Requested in #624 and #976.

**Improve rewind interval calculation**
This adds a minimum rewind interval of 2 seconds, so that when smart rewind is enabled, playback will always rewind by at least 2 seconds. The only exceptions are if we have been paused for less than 2 seconds or if we are less than 2 seconds from the start of the chapter. In these cases we will skip backwards by either the amount of time paused, or to the beginning of the chapter, whichever is closest. This is also requested in #717 and https://github.com/TortugaPower/BookPlayer/issues/976#issuecomment-1756982635.

This changes the easing function to be ease-out instead of ease-in. Instead of an easing function that starts slowly, then quickly increases towards the end I think the opposite is more natural. The amount we forget over time quickly increases at first, then slows down as time goes on. So I think the easing function should follow this. Also IMHO quartic feels like a little bit better of a fit than cubic here.

Finally, this changes smart rewind threshold up from 10 to 60 minutes. This seems like a big change, but with the change of the easing function it's less extreme—being paused for just 10 mins still gets you 50% of the rewind time, 26 mins get you 90%. IMO 10 minutes was a bit too short of a threshold, as personally I would want to skip back further after e.g. 45 minutes of pause time than after 10. See attached graphs to get a feel for how this looks in practice.

| Before | After |
|--------|--------|
| desmos-graph(1)  | desmos-graph  |
|  | NB: This is with the rewind limit set to 30s (0.5 min); that value is now configurable. The notch at the beginning is the 2 second minimum. |

- I've tested in the simulator and everything seems to work as expected
- I'm not sure how to handle updating the translation strings, as I've changed `settings_smartrewind_description` string but only in English for now.

Updated settings screen:
Simulator Screenshot - iPhone 17 Pro - 2025-10-19 at 11 19 37

### DIFF
--- a/BookPlayerWatch/LocalPlayback/Player/PlayerManager.swift
+++ b/BookPlayerWatch/LocalPlayback/Player/PlayerManager.swift
@@ -815,17 +815,34 @@ extension PlayerManager {
       let timePassedLimited = min(max(timePassed, 0), Constants.SmartRewind.threshold)
 
       let delta = timePassedLimited / Constants.SmartRewind.threshold
+      let baseRewindTime = ease(delta) * getMaxInterval()
 
-      // Using a cubic curve to soften the rewind effect for lower values and strengthen it for higher
-      let rewindTime = pow(delta, 3) * PlayerManager.rewindInterval
+      // Always rewind by at least minRewind...
+      let rewindTimeMin = max(baseRewindTime, Constants.SmartRewind.minRewind)
 
-      // Don't rewind past the beginning of the chapter
+      // ...but don't rewind past the beginning of the chapter, or more than the amount of time passed
       let timeInChapter = item.currentTime - item.currentChapter.start
-      let rewindTimeLimited = min(rewindTime, timeInChapter)
+      let rewindTimeLimited = min(rewindTimeMin, timeInChapter, timePassed)
 
       let newPlayerTime = max(CMTimeGetSeconds(self.audioPlayer.currentTime()) - rewindTimeLimited, 0)
 
       self.audioPlayer.seek(to: CMTime(seconds: newPlayerTime, preferredTimescale: CMTimeScale(NSEC_PER_SEC)))
+    }
+
+    func getMaxInterval() -> TimeInterval {
+      let userSetInterval = UserDefaults.standard.double(forKey: Constants.UserDefaults.smartRewindMaxInterval)
+      // The interval may be 0 if the setting has never been changed
+      return if userSetInterval > 0 {
+        userSetInterval
+      } else {
+        30.0
+      }
+    }
+
+    // Quartic ease out, where 0 ≤ delta ≤ 1
+    // Increases quickly at first, then levels off closer to 1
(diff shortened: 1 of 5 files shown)


GianniCarlo:
@lyallcooper thanks for the PR! I think a better approach would be to add a new row for an adjustable max rewind time (in `SettingsPlayerControlsView` → `SmartRewindSectionView`), so it's a picker like with the rewind / forward settings (see screenshot), this way it won't be tied to the user's rewind interval, and it'll fit a larger part of users

Screenshot 2025-10-18 (6 56 55 AM)

We can store that in another user defaults setting, so in the SwiftUI view the `@AppStorage` will drive that easily like we do for `SkipIntervalsSectionView`. After that's added, we can replace the '30 seconds' with `%@` in the base `settings_smartrewind_description` translation, so in the footer we can pass in the formatted time of the seconds (see `TimeParser.formatDuration(_)`), I can handle updating the rest of the translations

Let me know if this makes sense @lyallcooper

BookPlayerWatch/LocalPlayback/Player/PlayerManager.swift
```diff
@@ -815,14 +815,35 @@ extension PlayerManager {
...
 
-      // Using a cubic curve to soften the rewind effect for lower values and strengthen it for higher
-      let rewindTime = pow(delta, 3) * Constants.SmartRewind.maxTime
+      // Always rewind by at least minRewind...
+      let rewindTimeMin = max(baseRewindTime, Constants.SmartRewind.minRewind)
 
-      let newPlayerTime = max(CMTimeGetSeconds(self.audioPlayer.currentTime()) - rewindTime, 0)
+      // ...but don't rewind past the beginning of the chapter, or more than the amount of time passed
+      let timeInChapter = item.currentTime - item.currentChapter.start
+      let rewindTimeLimited = min(rewindTimeMin, timeInChapter, timePassed)
```
GianniCarlo, line 825: sorry it took me a while to get back to the PR (I was working on #1384), I'm not 100% convinced of the 2 secs minimum rewind and the 60 mins duration update (the rest looks good), but that just could be inertia to new things, I'm running this branch on my phone to see how it feels and report back between sunday and monday

GianniCarlo:
feels good, let's see if we get any feedback about how it works, worst case we can always add another setting. Nevermind about the build, that's just the specific simulator not being available, I fixed that on my other PR

PR #1387, opened 2025-10-30
Release v5.14.0

GianniCarlo:
Improvements

* You can now customize the speed quick actions with your preferred speed values
* Smart Rewind (thanks @lyallcooper !):
  * Max interval is now customizable too
  * The behavior was updated, so you get more of the rewind interval sooner, rather than by the end of the time
  * Please reach out to support@bookplayer.app with any feedback regarding this update
* Automated local backups for the database file, so your data can be recovered in case the database file gets corrupted

Bugfixes

* Carplay: Fix waiting for the sync operation to finish before resuming playback

If you'd like to contribute with translations to other languages, please reach us at support@bookplayer.app


lyallcooper:
I think some of the translations are missing the `settings_smartrewind_max_interval_title` string that was just added FYI

PR #1427, opened 2025-12-30
Add double tap gesture to now playing controls

jamesm131:
## Summary
- Adds double tap (primary hand gesture) support for play/pause in the Watch app's Now Playing view
- This enables the watchOS 11+ double tap gesture to toggle playback when using the watch as a companion/remote control for the iPhone

## Approach
Applied the existing `applyPrimaryHandGesture()` modifier to the play/pause button in `NowPlayingMediaControlsView.swift`. This modifier was already being used in `PlayerControlsView.swift` for local playback mode, so this change extends the same functionality to companion mode.

The implementation gracefully degrades on watchOS versions below 11.0.

## Things to watch out for
- Requires watchOS 11.0+ for the gesture to work
- No behaviour change on older watchOS versions


GianniCarlo (approved):
Thanks for catching this @jamesm131 ! I'll include it in the next version 🙌

PR #1464, opened 2026-02-19
Replacing PlayerViewController with native SwiftUI View implementation

Hirobreak:
## Purpose

- Migrate the existing ViewController implementation to SwiftUI to simplify UI structure, improve state management, and reduce maintenance complexity. This change aligns the codebase with a modern, declarative architecture.

## Approach

- Keeping the overall logic that orbits around using the PlayerManager, swap the UIKit views for a more modern SwiftUI approach

- The player is now an FullScreen Overlay

## Screenshots

| Before | After |
|--------|-------|
|  IMG_6686 | IMG_6685 |

### DIFF
--- a/BookPlayer/Coordinators/MainCoordinator.swift
+++ b/BookPlayer/Coordinators/MainCoordinator.swift
@@ -191,17 +191,11 @@ class MainCoordinator: NSObject {
   }
 
   func showPlayer() {
-    if let hostingController = mainController as? UIHostingController<MainView> {
-      hostingController.rootView.showPlayer()
-    }
+    playerState.showPlayer = true
   }
   
   func hasPlayerShown() -> Bool {
-    if let hostingController = mainController as? UIHostingController<MainView> {
-      return hostingController.rootView.hasPlayerShown()
-    }
-    
-    return false
+    return playerState.isShowingPlayer
   }
 
   func processFiles(urls: [URL]) {
--- a/BookPlayer/Library/MiniPlayer/MiniPlayerAccessoryView.swift
+++ b/BookPlayer/Library/MiniPlayer/MiniPlayerAccessoryView.swift
@@ -44,7 +44,7 @@ struct MiniPlayerAccessoryView: View {
         .padding(.leading, 8)
 
         VStack(alignment: .leading) {
-          Text(verbatim: relativePath ?? "")
+          Text(verbatim: playerManager.currentItem?.title ?? "")
             .foregroundStyle(
               reduceTransparency ? Color.primary : theme.primaryColor
             )
--- a/BookPlayer/MainView.swift
+++ b/BookPlayer/MainView.swift
@@ -16,7 +16,6 @@ struct MainView: View {
   @State private var listState = ListStateManager()
   @StateObject private var theme = ThemeViewModel()
   @StateObject private var keyboardObserver = KeyboardObserver()
-  @State private var isShowingPlayer: Bool = false
   @Environment(\.libraryService) private var libraryService
   @Environment(\.playerState) private var playerState
   @Environment(\.syncService) private var syncService
@@ -34,7 +33,6 @@ struct MainView: View {
       Tab("library_title", systemImage: "books.vertical") {
         LibraryRootView(
           showSecondOnboarding: showSecondOnboarding,
-          showPlayer: showPlayer,
           showImport: showImport
         )
         .background {
@@ -85,7 +83,7 @@ struct MainView: View {
         MiniPlayerAccessoryView(relativePath: relativePath, showPlayer: showPlayer)
       }
     }
-    .fullScreenCover(isPresented: $isShowingPlayer) {
+    .fullScreenCover(isPresented: playerState.isShowingPlayerBinding) {
       PlayerView {
         PlayerViewModel(
           libraryService: libraryService,
@@ -105,6 +103,12 @@ struct MainView: View {
     .tint(theme.linkColor)
     .onChange(of: scheme) {
       ThemeManager.shared.checkSystemMode()
+    }
+    .onChange(of: playerState.showPlayer) {
+      if playerState.showPlayer {
+        showPlayer()
+        playerState.showPlayer = false
+      }
     }
     .onReceive(
       NotificationCenter.default.publisher(for: .accountUpdate, object: nil)
@@ -127,11 +131,11 @@ struct MainView: View {
   }
   
   func showPlayer() {
-    isShowingPlayer = true
+    playerState.isShowingPlayer = true
   }
   
   func hasPlayerShown() -> Bool {
-    return isShowingPlayer
+    return playerState.isShowingPlayer
   }
 
   func handleDrop(_ providers: [NSItemProvider]) {
--- a/BookPlayer/Player/ViewModels/PlayerViewModel.swift
+++ b/BookPlayer/Player/ViewModels/PlayerViewModel.swift
@@ -10,33 +10,18 @@ import BookPlayerKit
 import Combine
 import SwiftUI
 
-enum PlayerSheetStyle {
-  case chapters
-  case controls
-  case buttonFree
-  case bookmark
-  case sleep
-}
+enum PlayerSheetStyle: String, Identifiable {
+    case chapters
+    case controls
+    case buttonFree
+    case bookmark
+    case sleep
 
-@Observable
-class PlayerSheetData {
-  var style: PlayerSheetStyle?
-  var display = false
-  
-  func displaySheet(style: PlayerSheetStyle) {
-    self.style = style
-    self.display = true
-  }
-  
-  func hideSheet() {
-    self.display = false
-    self.style = nil
-  }
+    var id: String { self.rawValue }
 }
 
 @MainActor
 final class PlayerViewModel: ObservableObject {
-  @Published var playerSheetData = PlayerSheetData()
   @Published var progressData = ProgressData()
   @Published var isPlaying = false
   @Published var playbackSpeed: Float = 1.0
@@ -50,11 +35,14 @@ final class PlayerViewModel: ObservableObject {
   @Published var hasNextChapter = false
   @Published var hasPreviousChapter = false
   @Published var isShowingNote: Bool = false
+  @Published var sheetStyle: PlayerSheetStyle?
+  @Published var displaySheet = false
   
-  let libraryService: LibraryService
-  let playbackService: PlaybackService
-  let playerManager: PlayerManager
-  let syncService: SyncService
+  let libraryService: LibraryServiceProtocol
+  let playbackService: PlaybackServiceProtocol
+  let playerManager: PlayerManagerProtocol
+  let syncService: SyncServiceProtocol
+  
   private var chapterBeforeSliderValueChange: PlayableChapter?
   private let sharedDefaults: UserDefaults
   private var prefersChapterContext: Bool
@@ -124,9 +112,11 @@ final class PlayerViewModel: ObservableObject {
     self.prefersChapterContext = sharedDefaults.bool(forKey: Constants.UserDefaults.chapterContextEnabled)
     self.prefersRemainingTime = sharedDefaults.bool(forKey: Constants.UserDefaults.remainingTimeEnabled)
     self.sharedDefaults = sharedDefaults
-    
+  }
+  
+  func bindBookObservers() {
     bindBookPlayingProgressEvents()
-    bindBookProgressSubscribers()
+    bindNotificationSubscribers()
     bindBookSharedObservers()
   }
   
@@ -151,23 +141,53 @@ final class PlayerViewModel: ObservableObject {
       
       object.set(false, forKey: Constants.UserDefaults.updateProgress)
     }
+    
+    SleepTimer.shared.$state
+      .map { [weak self] state -> String? in
+        switch state {
+        case .off:
+          return nil
+        case .endOfChapter:
+          return "active_title".localized
+        case .countdown(let seconds):
+          return self?.durationFormatter.string(from: seconds)
+        }
+      }
+      .removeDuplicates()
+      .receive(on: DispatchQueue.main)
+      .eraseToAnyPublisher()
+      .sink { [weak self] toolbarDescription in
+        guard let self else { return }
+
+        if let timeFormatted = toolbarDescription {
+          sleepText = timeFormatted
+        } else {
+          sleepText = nil
+        }
+      }.store(in: &disposeBag)
   }
   
-  func bindBookProgressSubscribers() {
-    self.playingProgressSubscriber?.cancel()
-    self.playingProgressSubscriber = NotificationCenter.default.publisher(for: .bookPlaying)
+  func bindNotificationSubscribers() {
+    NotificationCenter.default.publisher(for: .requestReview)
+      .debounce(for: 1.0, scheduler: DispatchQueue.main)
       .sink { [weak self] _ in
-        guard let self = self else { return }
-        self.recalculateProgress()
+        self?.requestReview()
       }
+      .store(in: &disposeBag)
+
+    NotificationCenter.default.publisher(for: .bookEnd)
+      .debounce(for: 1.0, scheduler: DispatchQueue.main)
+      .sink { [weak self] _ in
+        self?.requestReview()
+      }
+      .store(in: &disposeBag)
     
-    self.listeningProgressSubscriber?.cancel()
-    self.listeningProgressSubscriber = NotificationCenter.default.publisher(for: .listeningProgressChanged)
-      .receive(on: DispatchQueue.main)
+    NotificationCenter.default.publisher(for: UIDevice.batteryStateDidChangeNotification)
+      .debounce(for: 1.0, scheduler: DispatchQueue.main)
       .sink { [weak self] _ in
-        guard let self = self else { return }
-        self.recalculateProgress()
+        self?.handleAutolockStatus()
       }
+      .store(in: &disposeBag)
   }
   
   func bindBookPlayingProgressEvents() {
@@ -202,28 +222,21 @@ final class PlayerViewModel: ObservableObject {
       }
       .store(in: &disposeBag)
     
-    SleepTimer.shared.$state
-      .map { [weak self] state -> String? in
-        switch state {
-        case .off:
-          return nil
-        case .endOfChapter:
-          return "active_title".localized
-        case .countdown(let seconds):
-          return self?.durationFormatter.string(from: seconds)
-        }
-      }
+    self.playingProgressSubscriber?.cancel()
+    self.playingProgressSubscriber = NotificationCenter.default.publisher(for: .bookPlaying)
       .receive(on: DispatchQueue.main)
-      .eraseToAnyPublisher()
-      .sink { [weak self] toolbarDescription in
-        guard let self else { return }
-
-        if let timeFormatted = toolbarDescription {
-          sleepText = timeFormatted
-        } else {
-          sleepText = nil
-        }
-      }.store(in: &disposeBag)
+      .sink { [weak self] _ in
+        guard let self = self else { return }
+        self.recalculateProgress()
+      }
+    
+    self.listeningProgressSubscriber?.cancel()
+    self.listeningProgressSubscriber = NotificationCenter.default.publisher(for: .listeningProgressChanged)
+      .receive(on: DispatchQueue.main)
+      .sink { [weak self] _ in
+        guard let self = self else { return }
+        self.recalculateProgress()
+      }
   }
   
   private lazy var durationFormatter: DateComponentsFormatter = {
@@ -253,6 +266,16 @@ final class PlayerViewModel: ObservableObject {
         
         self?.relativePath = relativePath
       }
+  }
+  
+  func displaySheet(style: PlayerSheetStyle) {
+    self.sheetStyle = style
+    self.displaySheet = true
+  }
+  
+  func hideSheet() {
+    self.displaySheet = false
+    self.sheetStyle = nil
   }
   
   func getBookCurrentTime() -> TimeInterval {
@@ -327,12 +350,34 @@ final class PlayerViewModel: ObservableObject {
     self.chapterBeforeSliderValueChange = currentItem?.currentChapter
     
     progressData.chapterTitle = currentItem?.currentChapter?.title
-    ?? currentItem?.title
-    ?? ""
+      ?? currentItem?.title
+      ?? ""
     progressData.progress = progress
     progressData.maxTime = maxTimeInContext
     progressData.currentTime = currentTime
     progressData.sliderValue = Double(sliderValue)
+  }
+  
+  func handleNextTap() {
+    if let currentChapter = self.playerManager.currentItem?.currentChapter,
+      let nextChapter = self.playerManager.currentItem?.nextChapter(after: currentChapter)
+    {
+      self.playerManager.jumpToChapter(nextChapter)
+    } else {
+      self.playerManager.playNextItem(autoPlayed: false, shouldAutoplay: true)
+    }
+    NotificationCenter.default.post(name: .listeningProgressChanged, object: nil)
+  }
+  
+  func handlePreviousTap() {
+    if let currentChapter = self.playerManager.currentItem?.currentChapter,
+      let previousChapter = self.playerManager.currentItem?.previousChapter(before: currentChapter)
+    {
+      self.playerManager.jumpToChapter(previousChapter)
+    } else {
+      self.playerManager.playPreviousItem()
+    }
+    NotificationCenter.default.post(name: .listeningProgressChanged, object: nil)
   }
   
   func hasLoadedBook() -> Bool {
@@ -347,12 +392,12 @@ final class PlayerViewModel: ObservableObject {
         setMoreAlert()
       case .chapters:
         if UserDefaults.standard.bool(forKey: Constants.UserDefaults.playerListPrefersBookmarks) {
-          playerSheetData.displaySheet(style: .bookmark)
+          displaySheet(style: .bookmark)
         } else {
-          playerSheetData.displaySheet(style: .chapters)
+          displaySheet(style: .chapters)
         }
       case .speed:
-        playerSheetData.displaySheet(style: .controls)
+        displaySheet(style: .controls)
       case .timer:
         self.setSleepTimerAlert()
     }
@@ -402,7 +447,7 @@ final class PlayerViewModel: ObservableObject {
       BPActionItem(
         title: "sleeptimer_option_custom".localized,
         handler: { [weak self] in
-          self?.playerSheetData.displaySheet(style: .sleep)
+          self?.displaySheet(style: .sleep)
         }
       )
     )
@@ -455,17 +500,57 @@ final class PlayerViewModel: ObservableObject {
     )
   }
   
+  func requestReview() {
+    // don't do anything if flag isn't true
+    guard UserDefaults.standard.bool(forKey: "ask_review") else { return }
+    
+    // request for review if app is active
+    guard UIApplication.shared.applicationState == .active else { return }
+    
+#if RELEASE
+    AppDelegate.shared?.requestReview()
+#endif
+    
+    UserDefaults.standard.set(false, forKey: "ask_review")
+  }
+  
+  func handleAutolockStatus(forceDisable: Bool = false) {
+    guard !forceDisable else {
+      UIApplication.shared.isIdleTimerDisabled = false
+      UIDevice.current.isBatteryMonitoringEnabled = false
+      return
+    }
+    
+    guard UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabled) else {
+      UIApplication.shared.isIdleTimerDisabled = false
+      UIDevice.current.isBatteryMonitoringEnabled = false
+      return
+    }
+    
+    guard UserDefaults.standard.bool(forKey: Constants.UserDefaults.autolockDisabledOnlyWhenPowered) else {
+      UIApplication.shared.isIdleTimerDisabled = true
+      UIDevice.current.isBatteryMonitoringEnabled = false
+      return
+    }
+    
+    if !UIDevice.current.isBatteryMonitoringEnabled {
+      UIDevice.current.isBatteryMonitoringEnabled = true
+    }
+    
+    UIApplication.shared.isIdleTimerDisabled = UIDevice.current.batteryState != .unplugged
+  }
+  
   func resetShowings() {
-    playerSheetData.hideSheet()
+    hideSheet()
     isShowingNote = false
   }
   
   func showListFromMoreAction() {
     resetShowings()
     if UserDefaults.standard.bool(forKey: Constants.UserDefaults.playerListPrefersBookmarks) {
-      playerSheetData.displaySheet(style: .chapters)
+      displaySheet(style: .chapters)
     } else {
-      playerSheetData.displaySheet(style: .bookmark)
+      displaySheet(style: .bookmark)
     }
   }
   
@@ -481,28 +566,28 @@ final class PlayerViewModel: ObservableObject {
     actions.append(
       BPActionItem(
         title: self.getListTitleForMoreAction(),
-        handler: { self.showListFromMoreAction() }
+        handler: { [weak self] in self?.showListFromMoreAction() }
       )
     )
     
     actions.append(
       BPActionItem(
         title: "jump_start_title".localized,
-        handler: { self.handleJumpToStart() }
+        handler: { [weak self] in self?.handleJumpToStart() }
       )
     )
     
     actions.append(
       BPActionItem(
         title: markTitle,
-        handler: { self.handleMarkCompletion() }
+        handler: { [weak self] in self?.handleMarkCompletion() }
       )
     )
     
     actions.append(
       BPActionItem(
         title: "button_free_title".localized,
-        handler: { self.showButtonFree() }
+        handler: { [weak self] in self?.showButtonFree() }
       )
     )
     
@@ -510,7 +595,7 @@ final class PlayerViewModel: ObservableObject {
       BPActionItem(
         title: self.isRepeatEnabled()
           ? "repeat_turn_off_title".localized : "repeat_turn_on_title".localized,
-        handler: { self.handleEnableRepeat() }
+        handler: { [weak self] in self?.handleEnableRepeat() }
       )
     )
     
@@ -524,7 +609,7 @@ final class PlayerViewModel: ObservableObject {
   }
   
   func showButtonFree() {
-    playerSheetData.displaySheet(style: .buttonFree)
+    displaySheet(style: .buttonFree)
   }
   
   func createBookmark() {
@@ -568,10 +653,10 @@ final class PlayerViewModel: ObservableObject {
       actions.append(
         BPActionItem(
           title: "bookmark_note_action_title".localized,
-          handler: {
-            self.lastBookmark = bookmark
-            self.resetShowings()
-            self.isShowingNote = true
+          handler: { [weak self] in
+            self?.lastBookmark = bookmark
+            self?.resetShowings()
+            self?.isShowingNote = true
           }
         )
       )
@@ -580,8 +665,8 @@ final class PlayerViewModel: ObservableObject {
     actions.append(
       BPActionItem(
         title: "bookmarks_see_title".localized,
-        handler: {
-          self.playerSheetData.displaySheet(style: .bookmark)
+        handler: { [weak self] in
+          self?.displaySheet(style: .bookmark)
         }
       )
     )
@@ -590,8 +675,8 @@ final class PlayerViewModel: ObservableObject {
       BPActionItem(
         title: "ok_button".localized,
         style: .cancel,
-        handler: {
-          self.currentAlert = nil
+        handler: { [weak self] in
+          self?.currentAlert = nil
         }
       )
     )
--- a/BookPlayer/Player/Views/ArtworkView.swift
+++ b/BookPlayer/Player/Views/ArtworkView.swift
@@ -31,6 +31,7 @@ struct ArtworkView: View {
             KFImage
               .dataProvider(ArtworkService.getArtworkProvider(for: imagePath ?? ""))
               .targetCache(ArtworkService.cache)
+              .downsampling(size: CGSize(width: 100, height: 100))
               .resizable()
               .scaledToFill() // Fill the entire square
               .blur(radius: 20) // Adjust for more/less blur
@@ -73,7 +74,6 @@ struct ArtworkView: View {
                 }
               )
               .frame(maxWidth: .infinity, maxHeight: .infinity)
-              .background(Color.green)
           }
           .targetCache(ArtworkService.cache)
           .resizable()
--- a/BookPlayer/Player/Views/BPInputAlertModifier.swift
+++ b/BookPlayer/Player/Views/BPInputAlertModifier.swift
@@ -15,6 +15,7 @@ struct BPInputAlertContent: View {
   
   @State private var text = ""
   @FocusState private var focused: Bool
+  @AccessibilityFocusState private var isTextfieldFocused: Bool
   
   var body: some View {
     VStack(spacing: 0) {
@@ -26,7 +27,12 @@ struct BPInputAlertContent: View {
           .multilineTextAlignment(.center)
         
         TextField("bookmark_note_action_title".localized, text: $text)
-          .textFieldStyle(.roundedBorder)
+          .textFieldStyle(.plain) // Removes the default border
+          .padding(12) // Space between text and background
+          .background(
+              RoundedRectangle(cornerRadius: 10)
+                  .fill(Color(.systemGray6))
+          )
           .focused($focused)
       }
       .padding(20)
@@ -54,11 +60,14 @@ struct BPInputAlertContent: View {
       .fixedSize(horizontal: false, vertical: true)
       
     }
-    .frame(width: 270)
+    .frame(maxWidth: 270)
     .background(.regularMaterial)
     .clipShape(RoundedRectangle(cornerRadius: 14))
     .shadow(radius: 20)
     .onAppear { focused = true }
+    .accessibilityAction(.escape) { onCancel() }
+    .accessibilityAddTraits(.isModal)
+    .accessibilityFocused($isTextfieldFocused)
   }
 }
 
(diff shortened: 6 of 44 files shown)


BookPlayer/Player/Views/NewPlayerView.swift
```diff
@@ -0,0 +1,191 @@
...
+//  BookPlayer
+//
+//  Created by Pedro Iñiguez on 9/2/26.
+//  Copyright © 2026 BookPlayer LLC. All rights reserved.
+//
+
+import SwiftUI
+import BookPlayerKit
+
+struct NewPlayerView: View {
```
GianniCarlo: let's rename this to just `PlayerView`

GianniCarlo: apply the same to any 'New...' name

BookPlayer/Player/Views/NewPlayerView.swift
```diff
@@ -0,0 +1,191 @@
...
+    
+    if gesture.translation.height > threshold {
+      dismiss()
+    } else {
+      dragOffset = .zero
+    }
+  }
+}
+
+/*
```
GianniCarlo: remove this if previews won't be supported

BookPlayer/Player/Views/ListeningProgressView.swift
```diff
@@ -0,0 +1,64 @@
...
+        Spacer()
+        Button {
+          onRemainingToggle?()
+        } label: {
+          Text(remainingTime)
+            .frame(width: 60, alignment: .trailing)
+        }
+        .accessibilityLabel(remainingTimeAccessLabel)
+      }
+      .font(.caption)
```
GianniCarlo: use the fonts from `BPFont`

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,196 @@
...
+    .onAppear {
+      viewModel.recalculateProgress()
+    }
+    .accessibilityAction(.escape) {
+      dismiss()
+    }
+    .environmentObject(theme)
+    .onChange(of: scheme) {
+      ThemeManager.shared.checkSystemMode()
+    }
```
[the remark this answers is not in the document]
Hirobreak: `onChange(of:perform:)`was deprecated in iOS 17.0: Use `onChange` with a two or **zero** parameter action closure instead.

BookPlayer/Library/MiniPlayer/MiniPlayerAccessoryView.swift
```diff
@@ -44,7 +44,7 @@ struct MiniPlayerAccessoryView: View {
         .padding(.leading, 8)
 
         VStack(alignment: .leading) {
-          Text(verbatim: playerManager.currentItem?.title ?? "")
+          Text(verbatim: relativePath ?? "")
```
GianniCarlo: This would display a raw file path string (e.g. `"Books/MyBook.m4b"`) instead of the book's actual title. This would only affect users on iOS 26.0, on any other version this mini player view wouldn't be used

BookPlayer/Player/ViewModels/PlayerViewModel.swift
```diff
@@ -0,0 +1,619 @@
...
+  @Published var relativePath: String?
+  @Published var artworkUrl: URL?
+  @Published var currentAlert: BPAlertContent?
+  @Published var currentAlertOrigin: MediaAction?
+  @Published var sleepText: String?
+  @Published var hasNextChapter = false
+  @Published var hasPreviousChapter = false
+  @Published var isShowingNote: Bool = false
+  
+  let libraryService: LibraryService
```
GianniCarlo: The old ViewModel used protocol types (`PlayerManagerProtocol`, `LibraryServiceProtocol`, `SyncServiceProtocol`) which allowed injecting mocks for testing. See if we can keep using the protocols so we could add unit tests in the future

BookPlayer/Player/ViewModels/PlayerViewModel.swift
```diff
@@ -0,0 +1,619 @@
...
+    let markTitle = self.isBookFinished()
+      ? "mark_unfinished_title".localized
+      : "mark_finished_title".localized
+    
+    var actions = [BPActionItem]()
+
+    actions.append(
+      BPActionItem(
+        title: self.getListTitleForMoreAction(),
+        handler: { self.showListFromMoreAction() }
```
GianniCarlo: You're capturing `self` strongly. Since these closures are stored in `BPActionItem` objects held by `currentAlert` (a `@Published` property on `self`), this creates a retain cycle:

`ViewModel → currentAlert → BPActionItem.handler → ViewModel`

Use `[weak self]` in these closures:
```swift
handler: { [weak self] in self?.showListFromMoreAction() }
```

Same issue exists in `showBookmarkSuccessAlert()`

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,217 @@
...
+            DragGesture(minimumDistance: 15)
+                .onChanged(handleDragChanged)
+                .onEnded(handleDragEnded)
+        )
+    )
+    .contentShape(Rectangle())
+    .offset(y: dragOffset.height)
+    .animation(.interactiveSpring(), value: dragOffset)
+    .onAppear {
+      viewModel.recalculateProgress()
```
GianniCarlo, line 119: We're missing handling the autolock setting, that `PlayerCoordinator` had which:
- Disabled `UIApplication.shared.isIdleTimerDisabled` based on user settings
- Observed battery state changes to conditionally re-enable
- Re-enabled the idle timer on deallocation

The player screen will no longer prevent device sleep during playback.

Also missing: the old `PlayerViewController` observed `.requestReview` and `.bookEnd` notifications to trigger `SKStoreReviewController.requestReview()`

BookPlayer/Player/Views/ArtworkView.swift
```diff
@@ -0,0 +1,111 @@
...
+                      .padding(.vertical, 4)
+                      .accessibilityHidden(true)
+                  }
+                  .padding()
+                  .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .bottomLeading)
+                  .accessibilityHidden(true)
+                }
+              )
+              .frame(maxWidth: .infinity, maxHeight: .infinity)
+              .background(Color.green)
```
GianniCarlo: This looks like a debug leftover?

BookPlayer/Player/Views/BubbleButton.swift
```diff
@@ -0,0 +1,51 @@
...
+        if let text = labelText {
+          Text(text)
+            .bpFont(.headline)
+            .foregroundColor(theme.primaryColor)
+        }
+      }
+        .padding(12)
+        .frame(minWidth: 48)        // 2. Sets your "preferred" minimum width
+        .frame(height: 48)
+        .liquidGlassBackground()
```
GianniCarlo, line 43: Double check this works correctly and doesn't crash when compiling for a iOS 18 device

BookPlayer/Player/Views/ListeningProgressView.swift
```diff
@@ -0,0 +1,74 @@
...
+struct ListeningProgressView: View {
+  @EnvironmentObject private var theme: ThemeViewModel
+  @Binding var progress: Double
+  var remainingTime: String
+  var remainingTimeAccessLabel: String = ""
+  var currentTime: String
+  var currentTimeAccessLabel: String = ""
+  var progressLabel: String
+  var onSliderChange: ((Double) -> Void)?
+  var onProgresToggle: (() -> Void)?
```
GianniCarlo: typo `onProgresToggle` is missing an 's'. Should be `onProgressToggle`.

BookPlayer/Player/Views/DurationPickerSheet.swift
```diff
@@ -0,0 +1,108 @@
...
+    
+    var body: some View {
+        HStack(spacing: 0) {
+            // Hours Picker
+            Picker("Hours", selection: Binding(
+                get: { self.hours },
+                set: { newValue in updateDuration(newHours: newValue, newMinutes: self.minutes) }
+            )) {
+                ForEach(0..<24, id: \.self) { i in
+                    Text("\(i) hours").tag(i)
```
GianniCarlo: `"hours"` and `"min"` are hard-coded in English

BookPlayer/Player/Views/NavigationRowView.swift
```diff
@@ -0,0 +1,81 @@
...
+  var onTitleToggle: (() -> Void)?
+  
+  var body: some View {
+    HStack(spacing: 4) {
+      Button {
+        UIImpactFeedbackGenerator(style: .medium).impactOccurred()
+        if let currentChapter = self.playerManager.currentItem?.currentChapter,
+          let previousChapter = self.playerManager.currentItem?.previousChapter(before: currentChapter)
+        {
+          self.playerManager.jumpToChapter(previousChapter)
```
GianniCarlo: This view uses `playerManager` directly instead of going through the ViewModel. The old implementation centralized all playback actions through the ViewModel.

This means:
- Chapter navigation logic is duplicated here (was in ViewModel before)
- This view can't be previewed/tested without providing a real `PlayerManager`
- Some actions go through `PlayerViewModel`, others bypass it entirely

Consider passing action closures from the ViewModel instead, similar to `onTitleToggle`.

BookPlayer/Coordinators/MainCoordinator.swift
```diff
@@ -193,17 +191,17 @@ class MainCoordinator: NSObject {
...
   func showPlayer() {
-    let playerCoordinator = PlayerCoordinator(
-      flow: .modalOnlyFlow(presentingController: mainController!, modalPresentationStyle: .overFullScreen),
-      playerManager: self.playerManager,
-      libraryService: self.libraryService,
-      syncService: self.syncService
-    )
-    playerCoordinator.start()
+    if let hostingController = mainController as? UIHostingController<MainView> {
+      hostingController.rootView.showPlayer()
```
GianniCarlo: **Critical — Player will never appear from this code path.**

`hostingController.rootView` returns a *copy* of the `MainView` struct. Calling `.showPlayer()` on it sets `isShowingPlayer = true` on that copy, but `@State` storage is managed by SwiftUI's internal storage tied to the view's identity in the render tree — not to the struct instance. This mutation is discarded.

Same issue with `hasPlayerShown()` below — it reads from a fresh struct copy and will always return `false`.

This means the player won't appear when triggered from `AppDelegate` (e.g., remote commands, CarPlay, widget deep links).

**Suggested fix:** Use a shared `@Observable` or `ObservableObject` that both the coordinator and `MainView` reference. The existing `PlayerState` (already in the environment as `@Environment(\.playerState)`) could hold the `isShowingPlayer` flag.

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,217 @@
...
+    }
+    .accessibilityAction(.escape) {
+      dismiss()
+    }
+    .environmentObject(theme)
```
GianniCarlo, line 124: **Critical — Missing `playerManager` environment object.**

`.fullScreenCover` creates a **separate view hierarchy** that does NOT inherit `@EnvironmentObject` values from the presenting view. Only `theme` is injected here.

`PlayControlsRowView` and `NavigationRowView` both declare `@EnvironmentObject private var playerManager: PlayerManager` — they will crash at runtime with:
> "No ObservableObject of type PlayerManager found."

Need to add:
```swift
.environmentObject(theme)
.environmentObject(viewModel.playerManager)  // add this
```

Same issue inside the `.sheet` content (line 198) — if any sheet-presented sub-views also need `playerManager` as `@EnvironmentObject`, they'll crash too.

BookPlayer/Player/ViewModels/PlayerViewModel.swift
```diff
@@ -0,0 +1,619 @@
...
+  
+  func hideSheet() {
+    self.display = false
+    self.style = nil
+  }
+}
+
+@MainActor
+final class PlayerViewModel: ObservableObject {
+  @Published var playerSheetData = PlayerSheetData()
```
GianniCarlo: **Critical — `@Observable` inside `@Published` doesn't compose correctly for bindings.**

`PlayerSheetData` is `@Observable`, but it's stored as a `@Published` property on an `ObservableObject`. The `@Published` wrapper only fires `objectWillChange` when the *reference* changes (i.e., `playerSheetData = PlayerSheetData()`), NOT when internal properties like `.display` or `.style` are mutated.

This means `$viewModel.playerSheetData.display` (used in `.sheet(isPresented:)` on line 159 of PlayerView) relies on `ObservableObject`'s projected value for the binding, but mutating `display` won't fire `objectWillChange`. **The sheet may silently fail to present.**

Same concern applies to `progressData` on the next line — though the `@Bindable` usage in PlayerView's body does wire up `@Observable` tracking for the slider, other access patterns through `$viewModel.progressData` won't work.

**Fix options:**
1. Make `PlayerSheetData` and `ProgressData` plain structs (so `@Published` detects value changes)
2. Migrate `PlayerViewModel` entirely to `@Observable` (drop `ObservableObject`/`@Published`)
3. Use `sheet(item:)` with an `Identifiable` sheet style instead of `isPresented:`

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,217 @@
...
+      ThemeManager.shared.checkSystemMode()
+    }
+    .bpAlert($viewModel.currentAlert)
+    .bpInputAlert(
+        isPresented: $viewModel.isShowingNote,
+        title: "bookmark_note_action_title".localized
+    ) { text in
+      self.viewModel.saveNote(note: text)
+    }
+    .sheet(isPresented: $viewModel.playerSheetData.display) {
```
GianniCarlo: **High — Sheet transitions between styles will break.**

When `showListFromMoreAction()` calls `resetShowings()` (sets `display = false`) then immediately `displaySheet(style:)` (sets `display = true`), SwiftUI batches these in the same transaction and sees: `display` was `true`, still `true` → no-op. The sheet won't dismiss and re-present.

Additionally, `presentationDetents([.medium])` is only on the `.controls` case. If the sheet was first shown as `.chapters` (`.large`), switching to `.controls` won't retroactively apply the `.medium` detent.

**Recommended fix:** Use `.sheet(item:)` instead of `.sheet(isPresented:)`. Make `PlayerSheetStyle` conform to `Identifiable` and bind directly to an optional:
```swift
.sheet(item: $viewModel.currentSheetStyle) { style in
  switch style { ... }
}
```
This way SwiftUI auto-dismisses the old sheet and presents a new one when the item identity changes.

BookPlayer/Player/ViewModels/PlayerViewModel.swift
```diff
@@ -0,0 +1,619 @@
...
+      .receive(on: DispatchQueue.main)
+      .sink { [weak self] speed in
+        guard let self = self else { return }
+
+        self.playbackSpeed = speed
+        self.recalculateProgress()
+      }
+      .store(in: &disposeBag)
+    
+    SleepTimer.shared.$state
```
GianniCarlo: **High — Full body re-evaluation every second during sleep timer.**

`SleepTimer.shared.$state` publishes every second during countdown. Each tick updates `sleepText` (a `@Published` property), which triggers `objectWillChange` on the entire ViewModel, re-evaluating the **entire** `PlayerView.body` — artwork, slider, all buttons, all 5 dialog modifiers.

Unlike the `isPlaying`, `currentItem`, and `currentSpeed` subscribers, this one also lacks `.removeDuplicates()`, so even identical formatted strings trigger re-renders.

**Suggestions:**
1. Add `.removeDuplicates()` after the `.map`
2. Consider extracting the bottom toolbar into a separate child view with its own `@ObservedObject` or `@StateObject` so sleep timer ticks only redraw the toolbar, not the entire player

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,217 @@
...
+        .ignoresSafeArea()
+        .gesture(
+            DragGesture(minimumDistance: 15)
+                .onChanged(handleDragChanged)
+                .onEnded(handleDragEnded)
+        )
+    )
+    .contentShape(Rectangle())
+    .offset(y: dragOffset.height)
+    .animation(.interactiveSpring(), value: dragOffset)
```
GianniCarlo: **Medium — Animation scope is too broad + reduced motion not respected.**

`.animation(.interactiveSpring(), value: dragOffset)` is applied to the entire view tree. If a `@Published` change coincides with a drag offset change in the same transaction, unrelated property changes (sleep timer text, progress) could animate with the spring curve.

Also, `@Environment(\.accessibilityReduceMotion)` is not checked. Users with Reduce Motion enabled will still see spring animations.

**Suggestion:** Apply the animation more narrowly (on just the `.offset` modifier) and respect reduce motion:
```swift
.animation(reduceMotion ? .none : .interactiveSpring(), value: dragOffset)
```

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,217 @@
...
+  init(initModel: @escaping () -> PlayerViewModel) {
+    self._viewModel = .init(wrappedValue: initModel())
+  }
+  
+  var body: some View {
+    @Bindable var data = viewModel.progressData
+    
+    VStack {
+      DismissableRegionView()
+        .simultaneousGesture(
```
GianniCarlo, line 28: **High — Triple drag gesture can conflict with SlickSlider.**

Three `DragGesture(minimumDistance: 15)` are attached: on `DismissableRegionView` (`.simultaneousGesture`), `ArtworkView` (`.simultaneousGesture`), and the background `RoundedRectangle` (`.gesture`). The background covers the *entire* view including the slider area.

When the user drags the slider, the background's `DragGesture` will also try to recognize (it has the same minimum distance). This could cause the dismiss-drag to fight with the slider drag, especially on vertical swipes near the slider.

`DismissableRegionView` is also a `Button` — a quick tap-and-drag from the pill handle could trigger both the button tap (instant dismiss) and the drag gesture.

**Suggestion:** Use a single drag gesture at the container level with gesture priority, and use `.highPriorityGesture()` or coordinate space checks to prevent conflicts with the slider.

BookPlayer/Player/ViewModels/PlayerViewModel.swift
```diff
@@ -0,0 +1,619 @@
...
+        self.recalculateProgress()
+      }
+      
+      object.set(false, forKey: Constants.UserDefaults.updateProgress)
+    }
+  }
+  
+  func bindBookProgressSubscribers() {
+    self.playingProgressSubscriber?.cancel()
+    self.playingProgressSubscriber = NotificationCenter.default.publisher(for: .bookPlaying)
```
GianniCarlo: **High — Missing `.receive(on: DispatchQueue.main)`.**

The `.bookPlaying` notification subscriber here does NOT have `.receive(on: DispatchQueue.main)`, but the `.listeningProgressChanged` subscriber (line 165) does. If `.bookPlaying` is posted from a background thread, `recalculateProgress()` will be called off the main thread.

The class is `@MainActor`, but Combine's `sink` closure may not participate in actor isolation checking depending on the Swift concurrency mode. The inconsistency with the other subscriber suggests this was an oversight.

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,217 @@
...
+//
+
+import SwiftUI
+import BookPlayerKit
+
+struct PlayerView: View {
+  @Environment(\.colorScheme) private var scheme
+  @Environment(\.dismiss) private var dismiss
+  
+  @StateObject private var viewModel: PlayerViewModel
```
GianniCarlo, line 17: **Medium — `@StateObject` with eager init creates throwaway ViewModels.**

`StateObject(wrappedValue:)` evaluates its argument on every `init` call, but only *uses* the value from the first call. Since `PlayerView` is a struct inside a `.fullScreenCover` content closure, every time that closure evaluates, a new `PlayerViewModel` is constructed (with all its Combine subscriptions, KVO observers, and NotificationCenter subscriptions in `init`), and immediately discarded by SwiftUI.

These throwaway ViewModels briefly have active subscriptions before being deallocated.

**Suggestion:** Move subscription setup out of `init` into a method called from `.onAppear` or `.task`, so throwaway instances don't start subscriptions.

BookPlayer/Player/Views/BPInputAlertModifier.swift
```diff
@@ -0,0 +1,116 @@
...
+        .fontWeight(.semibold)
+      }
+      .fixedSize(horizontal: false, vertical: true)
+      
+    }
+    .frame(width: 270)
+    .background(.regularMaterial)
+    .clipShape(RoundedRectangle(cornerRadius: 14))
+    .shadow(radius: 20)
+    .onAppear { focused = true }
```
GianniCarlo, line 67: **Medium — Custom alert overlay has multiple accessibility issues.**

1. **No escape gesture** — VoiceOver users can't dismiss with two-finger scrub. Need `.accessibilityAction(.escape) { dismiss() }`
2. **No VoiceOver focus management** — `focused = true` sets *keyboard* focus, not VoiceOver focus. Use `.accessibilityFocused()` (iOS 15+) to move VoiceOver focus to the alert
3. **No modal trait** — VoiceOver users can navigate past the alert to underlying player controls. Add `.accessibilityAddTraits(.isModal)` to the alert container
4. **Dimmed background is a phantom a11y element** — `Color.black.opacity(0.35)` with `.onTapGesture` has no label. Add `.accessibilityHidden(true)` to it
5. **Fixed 270pt width** will clip text at large Dynamic Type sizes

BookPlayer/Player/Views/SlickSlider.swift
```diff
@@ -0,0 +1,124 @@
...
+      .frame(height: thumbSize)
+    }
+    .frame(height: thumbSize)
+    .onAppear { localValue = value }
+    .onChange(of: value) { _, newValue in
+        if !isDragging {
+            localValue = newValue
+        }
+    }
+    .accessibilityElement(children: .combine)
```
GianniCarlo, line 82: **Medium — Slider is missing `.accessibilityLabel`.**

The slider has `.accessibilityValue` (percentage) but no `.accessibilityLabel`. VoiceOver announces the value but not *what* the element represents. The old `ProgressSlider` had `accessibilityLabel` returning the percentage string.

Add something like:
```swift
.accessibilityLabel("player_scrubbing_title".localized)
```

Also note: VoiceOver won't re-announce the value during playback when the slider updates in real-time. Consider posting `UIAccessibility.post(notification: .layoutChanged)` when the value changes while focused.

BookPlayer/Player/Views/ArtworkView.swift
```diff
@@ -0,0 +1,111 @@
...
+        
+        Rectangle()
+          .fill(Color.secondary.opacity(0.1)) // A base color for while things load
+          .aspectRatio(1, contentMode: .fit)   // This locks it into a square
+          .overlay(
+            KFImage
+              .dataProvider(ArtworkService.getArtworkProvider(for: imagePath ?? ""))
+              .targetCache(ArtworkService.cache)
+              .resizable()
+              .scaledToFill() // Fill the entire square
```
GianniCarlo, line 36: **Medium — Performance: blur applied to full-resolution image.**

The blurred background loads the same image at full resolution with `.scaledToFill()` and applies `.blur(radius: 20)`. For a 1000x1000+ audiobook cover, this is expensive GPU work, and it re-runs on every `body` re-evaluation (which happens every second during sleep timer — see other comment).

Also, both `KFImage` instances use the same `ArtworkService.getArtworkProvider()`. DataProvider-based loading may not have Kingfisher's built-in request deduplication, potentially causing two disk reads.

**Suggestion:** For the blur background, use Kingfisher's `.downsampling(size:)` to load a small version (e.g., 100x100) — a low-res blurred image looks identical but is dramatically cheaper.

BookPlayer/Player/Views/NavigationRowView.swift
```diff
@@ -0,0 +1,81 @@
...
+//  BookPlayer
+//
+//  Created by Pedro Iñiguez on 9/2/26.
+//  Copyright © 2026 BookPlayer LLC. All rights reserved.
+//
+
+import SwiftUI
+
+struct NavigationRowView: View {
+  @EnvironmentObject private var playerManager: PlayerManager
```
GianniCarlo: **Medium — RTL layout will flip player controls.**

The old implementation set `.semanticContentAttribute = .forceLeftToRight` on player controls, following the Apple Books convention where media player controls are never flipped for RTL languages (rewind is always left, forward is always right).

The new SwiftUI implementation has no equivalent. In RTL locales, `HStack` will flip the layout — rewind button will appear on the right and forward on the left. This affects `NavigationRowView`, `PlayControlsRowView`, and `ListeningProgressView`.

**Fix:** Add `.environment(\.layoutDirection, .leftToRight)` to the relevant HStacks, or at the `PlayerView` level.

BookPlayer/Player/Views/SlickSlider.swift
Hirobreak: Tried this before put VoiceOver speaks twice, once for the value and a second for the label

BookPlayer/Player/Views/PlayControlsRowView.swift
```diff
@@ -0,0 +1,51 @@
...
+  @AppStorage(Constants.UserDefaults.forwardInterval) var forwardInterval: TimeInterval = 30
+  var isPlaying: Bool
+  @EnvironmentObject private var theme: ThemeViewModel
+  @EnvironmentObject private var playerManager: PlayerManager
+  
+  var body: some View {
+    HStack(spacing: 0) {
+      Spacer()
+      PlayerJumpView(backgroundImage: Image(.playerIconRewind), text: "-\(String(Int(rewindInterval.rounded())))", tintColor: Color(theme.linkColor)) {
+        UIImpactFeedbackGenerator(style: .medium).impactOccurred()
```
GianniCarlo, line 23: **Critical — Runtime crash: `playerManager` is never injected.**

`PlayControlsRowView` uses `@EnvironmentObject private var playerManager: PlayerManager`, but the player view hierarchy doesn't inject it anywhere.

The chain:
1. `MainView.fullScreenCover` → creates PlayerView (new view hierarchy, does NOT inherit `@EnvironmentObject`)
2. PlayerView `.environmentObject(theme)` → only theme is injected
3. PlayerView → embeds `PlayControlsRowView` → crashes with "No ObservableObject of type PlayerManager found"

Unlike `@Environment(\.playerManager)` (which uses `EnvironmentKey` and DOES propagate to fullScreenCover), `@EnvironmentObject` requires explicit injection.

**Fix options:**
1. Add `.environmentObject(viewModel.playerManager as! PlayerManager)` in PlayerView (near line 130)
2. Refactor PlayControlsRowView to accept action closures like NavigationRowView was refactored
3. Switch to `@Environment(\.playerManager)` if that key exists in the environment setup

Option 2 (closure callbacks) is cleanest and matches what was done for NavigationRowView.

BookPlayer/Player/Views/BubbleButton.swift
```diff
@@ -0,0 +1,51 @@
...
+        .clipShape(Capsule())
+    }
+  }
+}
+
+#Preview {
```
GianniCarlo, line 49: **High — `liquidGlassBackground()` is an iOS 26 API.**

This was flagged in the first review and is still present. If the project's minimum deployment target is lower than iOS 26, this will crash on older devices.

Please verify the deployment target or add:
```swift
if #available(iOS 26, *) {
  // use liquidGlassBackground()
} else {
  // fallback background style
}
```

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,193 @@
...
+    .animation(reduceMotion ? .none : .interactiveSpring(), value: dragOffset)
+    .onAppear {
+      viewModel.bindBookObservers()
+      viewModel.handleAutolockStatus(forceDisable: false)
+      viewModel.recalculateProgress()
+    }
+    .accessibilityAction(.escape) {
+      dismiss()
+    }
+    .environmentObject(theme)
```
GianniCarlo, line 124: **High — Missing `.onDisappear` to re-enable idle timer.**

The old `PlayerCoordinator` called `handleAutolockStatus(forceDisable: true)` in its `deinit` to re-enable the idle timer when the player was dismissed. The new implementation calls `handleAutolockStatus(forceDisable: false)` in `.onAppear` but never calls `handleAutolockStatus(forceDisable: true)` when the player is dismissed.

This means once the player is opened, the idle timer stays disabled even after returning to the library.

**Suggested fix:**
```swift
.onDisappear {
  viewModel.handleAutolockStatus(forceDisable: true)
}
```

BookPlayer/Player/ViewModels/PlayerViewModel.swift
```diff
@@ -0,0 +1,704 @@
...
+  }
+  
+  init(
+    libraryService: LibraryService,
+    playbackService: PlaybackService,
+    playerManager: PlayerManager,
+    syncService: SyncService
+  ) {
+    self.libraryService = libraryService
+    self.playbackService = playbackService
```
GianniCarlo, line 108: **High — `init` takes concrete types, preventing mock injection for testing.**

The stored properties are protocol-typed (lines 47-50), which is great. But the `init` parameters are concrete types (`LibraryService`, `PlaybackService`, `PlayerManager`, `SyncService`), so you can't construct a `PlayerViewModel` with mock implementations.

Since `PlayerViewModelTests` was deleted and the protocols are already declared as the property types, the fix is straightforward:
```swift
init(
  libraryService: LibraryServiceProtocol,
  playbackService: PlaybackServiceProtocol,
  playerManager: PlayerManagerProtocol,
  syncService: SyncServiceProtocol
) {
```

This would also allow restoring the deleted `PlayerViewModelTests`.

BookPlayer/Player/Views/SlickSlider.swift
```diff
@@ -0,0 +1,124 @@
...
+          .frame(width: thumbSize, height: thumbSize)
+          .shadow(color: accentColor.opacity(0.6), radius: 6)
+          .offset(x: horizontalBaseRatio * geometry.size.width - (thumbSize / 2))
+          .highPriorityGesture(
+            DragGesture(minimumDistance: 0)
+              .onChanged { gesture in
+                if !isDragging {
+                  isDragging = true
+                  onEditingChanged(true)
+                }
```
GianniCarlo, line 56: **Medium — Drag gesture still limited to the 18pt thumb.**

The `.highPriorityGesture` is only on the `Circle` thumb (18pt). Users must precisely hit this small target to start scrubbing. The old `ProgressSlider` (UISlider) allowed tapping/dragging anywhere on the track.

Consider adding a transparent overlay over the full track width to increase the hit area, or adding a tap gesture on the ZStack for tap-to-seek.

BookPlayer/Player/ViewModels/PlayerViewModel.swift
```diff
@@ -0,0 +1,704 @@
...
+  @Published var hasPreviousChapter = false
+  @Published var isShowingNote: Bool = false
+  @Published var sheetStyle: PlayerSheetStyle?
+  @Published var displaySheet = false
+  
+  let libraryService: LibraryServiceProtocol
+  let playbackService: PlaybackServiceProtocol
+  let playerManager: PlayerManagerProtocol
+  let syncService: SyncServiceProtocol
+
```
GianniCarlo, line 45: **Medium — `displaySheet` appears to be unused dead code.**

`@Published var displaySheet = false` is set in `displaySheet(style:)` and `hideSheet()`, but PlayerView uses `.sheet(item: $viewModel.sheetStyle)` — binding directly to `sheetStyle`. The `displaySheet` boolean isn't bound to anything in the view.

This looks like a leftover from the refactor when `PlayerSheetData` was removed. Consider removing it to avoid confusion.

GianniCarlo (commented):
## Functionality Review (v2)

Great progress since the first round! Many of the issues were addressed:
- Auto-lock/idle timer management restored
- requestReview/bookEnd notification handling restored
- Retain cycles in action closures fixed with `[weak self]`
- MiniPlayerAccessoryView title bug fixed
- DurationPickerSheet localized with `Measurement` formatter
- Typo `onProgresToggle` corrected
- `Color.green` debug background removed
- NavigationRowView refactored to use closure callbacks from ViewModel
- MediaActionRow extracted as separate component

Some issues remain — see inline comments.

### Summary of Remaining Findings

| Priority | Issue |
|----------|-------|
| **Critical** | `PlayControlsRowView` uses `@EnvironmentObject var playerManager` but it's never injected — will crash at runtime |
| **High** | `liquidGlassBackground()` — still needs iOS version availability check |
| **High** | Missing `.onDisappear` to re-enable idle timer when player dismisses |
| **High** | ViewModel `init` still takes concrete types — can't construct with mocks for testing |
| **Medium** | SlickSlider only draggable from 18pt thumb, not the full track |
| **Medium** | Unused `@Published var displaySheet` property (dead code from refactor) |
| **Low** | `Color.white.opacity(0.15)` hard-coded in SlickSlider background track |

BookPlayer/Player/Views/SlickSlider.swift
```diff
@@ -0,0 +1,124 @@
...
+      ZStack(alignment: .leading) {
+        // Background Track
+        Capsule()
+          .fill(Color.white.opacity(0.15))
+          .frame(height: trackHeight)
+        
+        // Active Track
+        Capsule()
+          .fill(accentColor)
+          .frame(width: max(0, horizontalBaseRatio * geometry.size.width), height: 4)
```
GianniCarlo, line 42: **Low — Hard-coded `Color.white.opacity(0.15)` won't be visible in light mode.**

This was noted in the first review. Consider using a theme-aware color like `Color.secondary.opacity(0.2)` or passing it as a parameter, similar to `accentColor`.

BookPlayer/MainView.swift
```diff
@@ -84,6 +83,17 @@ struct MainView: View {
...
+    .fullScreenCover(isPresented: playerState.isShowingPlayerBinding) {
+      PlayerView {
+        PlayerViewModel(
+          libraryService: libraryService,
+          playbackService: playbackService,
+          playerManager: playerManager,
+          syncService: syncService
+        )
+      }
+      .presentationBackground(.clear)
```
GianniCarlo, line 95: **Critical — `fullScreenCover` does NOT inherit `@EnvironmentObject` from parent.**

The `.fullScreenCover` creates a **separate view hierarchy**. Any `@EnvironmentObject` values (like `playerManager`) from the parent chain are NOT available inside.

Currently only `theme` is injected (in PlayerView). But `PlayControlsRowView` declares `@EnvironmentObject private var playerManager: PlayerManager` and will crash:
> "No ObservableObject of type PlayerManager found."

Note that `@Environment(\.playerManager)` (using `EnvironmentKey`) DOES propagate to fullScreenCover content. But `@EnvironmentObject` does not.

**Suggested fix:** Refactor `PlayControlsRowView` to use closure callbacks (like NavigationRowView was refactored), or inject playerManager here:
```swift
.fullScreenCover(isPresented: playerState.isShowingPlayerBinding) {
  PlayerView { ... }
  .environmentObject(playerManager)  // add this
  .presentationBackground(.clear)
}
```

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,193 @@
...
+  var body: some View {    
+    VStack {
+      DismissableRegionView()
+        .simultaneousGesture(
+          DragGesture(minimumDistance: 15)
+            .onChanged { gesture in
+              handleDragChanged(gesture)
+            }
+            .onEnded { gesture in
+              handleDragEnded(gesture)
```
GianniCarlo, line 34: **High — Three overlapping `DragGesture(minimumDistance: 15)` remain.**

Drag gestures are attached to:
1. `DismissableRegionView` (line 34, `.simultaneousGesture`)
2. `ArtworkView` (line 50, `.simultaneousGesture`)
3. Background `RoundedRectangle` (line 114, `.gesture`)

While SlickSlider now uses `.highPriorityGesture` (which helps), the three dismiss-drag gestures can still compete with each other and with the `DismissableRegionView` button tap (a quick tap-and-drag from the pill handle triggers both).

**Suggestion:** Consolidate into a single drag gesture at the container level. For example, place one `.gesture(DragGesture(...))` on the outermost VStack and remove the other two. The `.highPriorityGesture` on SlickSlider's thumb will still take priority for slider drags.

BookPlayer/Player/Views/BPInputAlertModifier.swift
```diff
@@ -0,0 +1,126 @@
...
+    .accessibilityAction(.escape) { onCancel() }
+    .accessibilityAddTraits(.isModal)
+    .accessibilityFocused($isTextfieldFocused)
+  }
+}
+
+struct BPInputAlertModifier: ViewModifier {
+  @Binding var isPresented: Bool
+
```
GianniCarlo, line 76: **High — `@AccessibilityFocusState` declared but never triggered.**

`isTextfieldFocused` (line 24) and `.accessibilityFocused($isTextfieldFocused)` (line 76) are wired up, but `isTextfieldFocused` is never set to `true`. The `.onAppear` on line 73 sets keyboard `focused = true` but not `isTextfieldFocused`.

VoiceOver focus won't move to the alert when it appears.

**Fix:**
```swift
.onAppear {
  focused = true
  isTextfieldFocused = true
}
```

BookPlayer/Player/Views/SlickSlider.swift
```diff
@@ -0,0 +1,124 @@
...
+            localValue = newValue
+        }
+    }
+    .accessibilityElement(children: .combine)
+    .accessibilityValue(getAccessibilityLabel(localValue))
+    .accessibilityAdjustableAction { direction in
+      isDragging = true
+      onEditingChanged(true)
+      
+      let step = (range.upperBound - range.lowerBound) * 0.05 // 5% jumps
```
GianniCarlo, line 88: **Medium — Slider is missing `.accessibilityLabel`.**

The slider has `.accessibilityValue` (percentage via `getAccessibilityLabel()`) and `.accessibilityAdjustableAction`, but no `.accessibilityLabel`. VoiceOver announces the percentage value but not *what* the element represents.

Add:
```swift
.accessibilityLabel("player_scrubbing_title".localized)
```

(or similar localized string like "Playback position")

BookPlayer/Player/Views/PlayerView.swift
```diff
@@ -0,0 +1,193 @@
...
+  @State private var dragOffset: CGSize = .zero
+  
+  init(initModel: @escaping () -> PlayerViewModel) {
+    self._viewModel = .init(wrappedValue: initModel())
+  }
+  
+  var body: some View {    
+    VStack {
+      DismissableRegionView()
+        .simultaneousGesture(
```
GianniCarlo, line 28: **Medium — `@StateObject` init still eagerly evaluates.**

`StateObject(wrappedValue:)` evaluates its argument on every `init` call. The `initModel()` closure is called immediately during `PlayerView.init`, meaning a new `PlayerViewModel` is constructed each time the closure evaluates (which happens whenever the fullScreenCover content closure runs).

The good news: subscriptions are now in `bindBookObservers()` (called from `.onAppear`), so throwaway ViewModels won't start Combine subscriptions. However, the `init` still sets up `UserDefaults.sharedDefaults`, `prefersChapterContext`, etc., and allocates the disposeBag. This is minor overhead but still creates unnecessary objects.

The standard pattern is:
```swift
_viewModel = StateObject(wrappedValue: initModel())
```
which is what you have — the issue is inherent to `StateObject`'s eager evaluation. If you want to avoid it entirely, consider using `@State` + `task` pattern available in newer SwiftUI, or `.task { viewModel.setup() }` for deferred init work.

BookPlayer/Player/ViewModels/PlayerViewModel.swift
```diff
@@ -0,0 +1,704 @@
...
+    isShowingNote = false
+  }
+  
+  func showListFromMoreAction() {
+    resetShowings()
+    if UserDefaults.standard.bool(forKey: Constants.UserDefaults.playerListPrefersBookmarks) {
+      displaySheet(style: .chapters)
+    } else {
+      displaySheet(style: .bookmark)
+    }
```
GianniCarlo, line 554: **Low — `showListFromMoreAction()` may cause a sheet flash.**

`resetShowings()` calls `hideSheet()` which sets `sheetStyle = nil`. Then immediately `displaySheet(style:)` sets `sheetStyle` to a new value. With `.sheet(item:)`, SwiftUI will dismiss the current sheet (identity changed) and present the new one.

Depending on SwiftUI's transaction batching, this could either:
- Work smoothly (if batched in same transaction)
- Cause a brief flash where the old sheet dismisses before the new one presents

If you see a flash, consider adding a small delay or using a single state update:
```swift
func showListFromMoreAction() {
  // Directly switch style without going through nil
  sheetStyle = UserDefaults.standard.bool(...) ? .chapters : .bookmark
}
```

GianniCarlo (commented):
## SwiftUI Best Practices Review (v2)

Second pass focused on SwiftUI-specific patterns. Great improvements since v1:
- `PlayerState` shared observable fixes the coordinator ↔ SwiftUI bridge
- `@Observable`/`@Published` mixing resolved — `ProgressData` is now a plain struct, `PlayerSheetData` removed
- `.sheet(item:)` replaces `.sheet(isPresented:)` — auto-handles sheet identity changes
- RTL layout forced LTR on player controls
- `accessibilityReduceMotion` respected
- BPInputAlertModifier: escape gesture, modal trait, and accessibility focus added
- `.removeDuplicates()` on sleep timer publisher
- `.receive(on: .main)` on `.bookPlaying` notification
- Blur uses `.downsampling(size:)` for performance
- Subscriptions moved to `bindBookObservers()` (called in `.onAppear`) — throwaway ViewModels no longer start subscriptions

### Remaining Findings

| # | Priority | Issue |
|---|----------|-------|
| 1 | **Critical** | `PlayControlsRowView` `@EnvironmentObject var playerManager` — not injected in fullScreenCover hierarchy |
| 2 | **High** | Triple overlapping drag gestures still present |
| 3 | **High** | `@AccessibilityFocusState` declared but never set to `true` |
| 4 | **Medium** | SlickSlider missing `.accessibilityLabel` |
| 5 | **Medium** | `@StateObject` init still eagerly evaluates — creates KVO observer in `init` |
| 6 | **Low** | `showListFromMoreAction()` calls `resetShowings()` then `displaySheet()` — sheet identity change may flash |

PR #1469, opened 2026-03-07
Library UUID

Hirobreak:
## Purpose

- Assign uuids to library items, this way it is easier to map actions independently of where the items are

## Approach

- Generate optional uuids to avoid breaking current db status
- Sync batches of library items, reporting their local uuid or reassigning the existing server-side uuid
- Newer actions will attach its uuid

- Persisting both relativePath and uuid for existing actions


Shared/CoreData/Backed-Models/LibraryItem+CoreDataProperties.swift
```diff
@@ -15,6 +15,7 @@ extension LibraryItem {
     return NSFetchRequest<LibraryItem>(entityName: "LibraryItem")
   }
 
+  @NSManaged public var uuid: String
```
GianniCarlo, line 18: **CRITICAL: CoreData optional vs Swift non-optional mismatch**

The CoreData schema (`Audiobook Player 11.xcdatamodel`) defines `uuid` as `optional="YES"`, but this Swift property is declared as non-optional `String`. If any item ends up with a nil UUID (e.g., a partial migration failure, or an item created through a code path that doesn't set it), accessing `.uuid` will crash at runtime.

Additionally, `SimpleLibraryItem.uuid` is also declared as non-optional (`let uuid: String`), and the `parseFetchedItems` uses `guard let uuid = dictionary["uuid"] as? String` — which means items without a UUID will be **silently filtered out from all list views** during the window between CoreData migration and matchUuid completion.

Consider either:
- Making the Swift property `String?` and updating callers, or
- Making the CoreData attribute non-optional with a default value expression (and ensure the migration policy guarantees a value is always set)

Shared/CoreData/Lightweight-Models/PlayableItem.swift
```diff
@@ -218,6 +222,7 @@ extension PlayableItem: Codable {
       currentTime: try values.decode(TimeInterval.self, forKey: .currentTime),
       duration: try values.decode(TimeInterval.self, forKey: .duration),
       relativePath: try values.decode(String.self, forKey: .relativePath),
+      uuid: try values.decode(String.self, forKey: .uuid),
```
GianniCarlo: **CRITICAL: Codable decode will crash on old persisted data**

`try values.decode(String.self, forKey: .uuid)` will throw if the key is missing. Old `PlayableItem` data encoded before this PR (e.g., via WatchConnectivity or UserDefaults) won't have a `"uuid"` key. Since the property type is `String?` (optional), this should use:
```swift
uuid: try values.decodeIfPresent(String.self, forKey: .uuid),
```
to maintain backward compatibility.

Shared/Services/Sync/SyncTasksStorage.swift
```diff
@@ -28,9 +28,10 @@ public actor SyncTasksStorage: ModelActor {
   public func appendTask(parameters: [String: Any]) async throws {
     guard
       let taskId = parameters["id"] as? String,
-      let relativePath = parameters["relativePath"] as? String,
       let rawJobType = parameters["jobType"] as? String,
-      let jobType = SyncJobType(rawValue: rawJobType)
+      let jobType = SyncJobType(rawValue: rawJobType),
+      let uuid = parameters["uuid"] as? String,
```
GianniCarlo, line 33: **CRITICAL: Sync tasks with nil uuid are silently dropped**

This `guard` now requires `uuid` to be present in `parameters`. However, several scheduling methods (`scheduleDeleteJob`, `scheduleDeleteBookmarkJob`, `scheduleSetBookmarkJob`, `scheduleRenameFolderJob`, `scheduleArtworkUpload`) only **conditionally** add uuid with `if let uuid { params["uuid"] = uuid }`. When uuid is nil (e.g., items that haven't synced UUIDs yet), this guard will fail and throw, silently dropping the sync task.

Either:
- Make uuid optional in this guard and handle nil downstream, or
- Ensure all callers always provide a uuid (even a fallback empty string)

Shared/Services/Sync/SyncJobScheduler.swift
```diff
@@ -351,22 +395,100 @@ public class SyncJobScheduler: JobSchedulerProtocol, BPLogger {
...
+        self.tasksProgress.removeValue(forKey: task.uuid)
+      }
+    }
+  }
+  
+  private func handleMatchUuidsResponse(_ results: MatchUuidsResponse) {
+    let context = dataManager.getBackgroundContext()
+    
+    context.perform { [unowned self, context] in
+      let uuidsToUpdate = results.conflicts.map { $0.key }
```
GianniCarlo: **CRITICAL: Conflict resolution predicate searches `uuid` using path keys**

`results.conflicts.map { $0.key }` returns file paths/names (e.g., `"Jim Butcher - Ghost Story - 17.mp3"`), but the NSPredicate below searches `uuid IN %@` with those values. Since `key` is a relativePath and not a UUID, this predicate will never match any items, making **conflict resolution non-functional**.

Similarly, the SwiftData portion below searches `SyncTaskReferenceModel` where `container.uuid == conflictUuid` (the NEW uuid from server), but existing task refs would still have the OLD uuid — so this won't match either.

The predicate should likely use `relativePath IN %@` for the CoreData fetch, and the SwiftData search should match on the old uuid (the `key`).

Shared/Services/Sync/SyncJobScheduler.swift
```diff
@@ -351,22 +395,100 @@ public class SyncJobScheduler: JobSchedulerProtocol, BPLogger {
...
+          item.uuid = newUUID
+        }
       }
+      
+      try? context.save()
     }
+    
+    let modelContext = ModelContext(tasksDataManager.container)
+    modelContext.autosaveEnabled = true
+
```
GianniCarlo: **CRITICAL: Threading issues with ModelContext outside actor isolation**

This `ModelContext` is created from `operationTask.completionBlock` which runs on an arbitrary thread. `ModelContext` is not thread-safe and should respect SwiftData's actor isolation.

Additional concerns in this method:
- The CoreData `context.perform` block uses `unowned self` which can crash if `SyncJobScheduler` is deallocated while the operation is in flight
- `try? context.save()` silently swallows CoreData save errors
- The CoreData and SwiftData saves are not coordinated — if one succeeds and the other fails, state becomes inconsistent (items have updated UUIDs in CoreData but SwiftData tasks still reference old ones, or vice versa)

Shared/SwiftData/MigrationPlan.swift
```diff
@@ -0,0 +1,97 @@
...
+    willMigrate: { _ in
+    },
+    didMigrate: { context in
+      guard let coreDataContext = injectedCoreDataContext else {
+        fatalError("Core Data context was not injected before migration!")
+      }
+      // 1. Fetch all V2 models (which have both path and optional uuid)
+      let items = try context.fetch(FetchDescriptor<SchemaV2.SyncTaskReferenceModel>())
+      for item in items {
+        item.uuid = "LEGACY_UUID_PLACEHOLDER"
```
GianniCarlo: **WARNING: Placeholder UUID may leak to API**

`"LEGACY_UUID_PLACEHOLDER"` is set for existing `SyncTaskReferenceModel` items during migration. If the sync service processes these tasks before `matchUuid` jobs complete, the API will receive this placeholder string instead of a real UUID. 

Consider either:
- Looking up the real UUID from CoreData during migration (similar to the matchUuid batch logic), or
- Adding a guard in the sync service to skip/defer tasks with this placeholder value

Shared/SwiftData/MigrationPlan.swift
```diff
@@ -0,0 +1,97 @@
...
+public enum MigrationPlan: SchemaMigrationPlan {
+  public static var schemas: [any VersionedSchema.Type] {
+    [SchemaV1.self, SchemaV2.self]
+  }
+  
+  public static var stages: [MigrationStage] {
+    [v1ToV2]
+  }
+  
+  public static var injectedCoreDataContext: NSManagedObjectContext?
```
GianniCarlo, line 21: **WARNING: Static mutable state for CoreData context injection**

This is set in `AppDelegate` but the `TasksDataManager` init (which triggers migration via `ModelContainer`) could theoretically run before that assignment if initialization order changes. The `fatalError` at line 31 would crash the app.

Also, `NSManagedObjectContext` is not thread-safe — the `backgroundContext` is a private queue context, yet it's accessed from whatever thread SwiftData runs its migration on. While `performAndWait` is used inside the migration, the assignment and guard check happen outside any queue synchronization.

Shared/Services/LibraryService+Sync.swift
```diff
@@ -45,9 +45,33 @@ public protocol LibrarySyncProtocol {
   func getBookmarks(of type: BookmarkType, relativePath: String) -> [SimpleBookmark]?
   /// Store new synced bookmark
   func addBookmark(from bookmark: SimpleBookmark) async
+  
+  func generateMissingUuids(offset: Int) async -> [String: String]
 }
 
 extension LibraryService: LibrarySyncProtocol {
+  public func generateMissingUuids(offset: Int) async -> [String: String] {
```
GianniCarlo: **WARNING: Fetches ALL items, not just ones with missing UUIDs**

The function is named `generateMissingUuids` but the fetch request has no predicate — it fetches all items. This means every call loops through the entire library, creating batch `matchUuid` sync tasks for ALL items (not just new/unmatched ones).

Consider adding a predicate like `NSPredicate(format: "uuid == nil OR uuid == ''")` if the intent is to only process items with missing UUIDs, or rename the function if fetching all is intentional.

Also: `[unowned self]` on line 56 risks a crash if `LibraryService` is deallocated before the background context block executes — `weak self` with guard would be safer.

Shared/SwiftData/SwiftDataModels+Extensions.swift
```diff
@@ -85,8 +87,8 @@ extension DeleteTaskModel: DictionaryConvertible {
   public func toDictionaryPayload() -> [String: Any] {
```
GianniCarlo, line 96: **WARNING: `toDictionaryPayload` drops `relativePath` for delete/bookmark/rename/artwork tasks**

This and several other `toDictionaryPayload` implementations (`DeleteBookmarkTaskModel`, `SetBookmarkTaskModel`, `RenameFolderTaskModel`, `ArtworkUploadTaskModel`) now emit `uuid` instead of `relativePath`. If the API hasn't been updated to accept `uuid` in place of `relativePath` for ALL these endpoints, these operations will break.

Please confirm the backend handles `uuid`-only payloads for all these job types.

Shared/SwiftData/Models/SchemaV2SyncTasksModels.swift
```diff
@@ -0,0 +1,281 @@
...
+  
+  /// SwiftData model with type and task id reference to the real task
+  @Model
+  public class SyncTaskReferenceModel {
+    @Attribute(.unique) public var id: String
+    public var relativePath: String
+    public var taskID: String
+    public var jobType: SyncJobType
+    public var position: Int
+    public var uuid: String = UUID().uuidString
```
GianniCarlo: **WARNING: Existing task models get random UUIDs during migration**

During V1→V2 migration, SwiftData uses this default `UUID().uuidString` for the new `uuid` column. This means **all existing pending task models** (Upload, Update, Move, Delete, etc.) will get random UUIDs that don't correspond to any actual library items.

Only `SyncTaskReferenceModel` is handled in `didMigrate` (set to placeholder). The `didMigrate` handler should also iterate over other V2 task models and populate their `uuid` from CoreData — otherwise existing pending tasks will carry meaningless random UUIDs.

Shared/Services/Sync/SyncService.swift
```diff
@@ -503,6 +520,22 @@ extension SyncService {
       await jobManager.scheduleMetadataUpdateJob(with: relativePath, parameters: params)
     }
   }
+  
+  public func scheduleMatchUuid() async {
```
GianniCarlo: **WARNING: No deduplication for matchUuid tasks**

This method loops through ALL items in batches of 200 and creates a separate sync task per batch. There's no check for already-queued `matchUuid` tasks — calling `scheduleMatchUuid()` twice queues everything twice.

Consider checking if matchUuid tasks are already queued before scheduling more, or clearing existing ones first.

Also: `uuidsDict.count > 0` on line 531 — idiomatic Swift prefers `!uuidsDict.isEmpty`.

Shared/Services/Sync/SyncJobScheduler.swift
```diff
@@ -286,58 +324,64 @@ public class SyncJobScheduler: JobSchedulerProtocol, BPLogger {
...
         }
-
+        
         let operationTask = LibraryItemSyncOperation(
           client: networkClient,
           task: task
         )
-
+        
         operationTask.completionBlock = { [unowned self, unowned operationTask] in
```
GianniCarlo: **WARNING: `unowned self` + `unowned operationTask` in completion block**

If `SyncJobScheduler` is deallocated while an operation is in flight (e.g., user logs out during sync), this will crash. The new `handleMatchUuidsResponse` call in this closure does significant work (CoreData + SwiftData operations), extending the window where a crash can occur.

Consider using `weak self` with a guard instead.

Shared/CoreData/Migrations/ManualOrderMigrationUtils.swift
```diff
@@ -41,4 +41,19 @@ extension DataMigrationManager {
 
     dataManager.saveContext()
   }
+  
+  func populateUuids(dataManager: DataManager) {
```
GianniCarlo: **WARNING: `populateUuids()` appears to be dead code**

This method is defined but never called anywhere. The `UUIDMigrationPolicy` already handles UUID generation during the heavyweight migration. Additionally, this method generates **different** UUIDs than the migration policy would — if it were ever called post-migration, it would overwrite the UUIDs assigned during migration.

Either remove this dead code, or if it's intended as a safety net, add it to `DataMigrationManager.performMigration()` as a `case .v10:` post-migration step with a `uuid == nil` predicate to avoid overwriting existing UUIDs.

BookPlayer/Library/ItemList/Models/LibraryNode.swift
```diff
@@ -10,20 +10,27 @@ import Foundation
...
     switch self {
     case .root, .book: nil
-    case .folder(_, let relativePath): relativePath
+    case .folder(_, let relativePath, _): relativePath
+    }
+  }
+  
+  var uuid: String {
+    switch self {
+    case .root: ""
```
GianniCarlo, line 32: **WARNING: Empty string for root uuid is a silent failure mode**

Returning `""` instead of `nil` (or making `uuid` optional) means callers may unknowingly pass an empty string to API calls that expect a real UUID. The `useUuids = itemOrigin.uuid != ""` check in `scheduleMoveItemJob` catches this for moves, but other call sites might not.

Shared/CoreData/CoreDataStack.swift
```diff
@@ -46,7 +46,7 @@ public class CoreDataStack {
     self.storeContainer = NSPersistentContainer(name: modelName)
 
     let description = NSPersistentStoreDescription()
-    description.shouldInferMappingModelAutomatically = false
+    description.shouldInferMappingModelAutomatically = true
```
GianniCarlo: **SUGGESTION:** This change is in the test initializer (`init(testPath:)`), so production is not affected. Just flagging for awareness — with `shouldInferMappingModelAutomatically = true` in tests, test migrations may behave differently than production (which uses explicit mapping models via `DataMigrationManager`).

BookPlayer/Library/ItemList/ItemListViewModel.swift
```diff
@@ -364,18 +368,18 @@ extension ItemListViewModel {
     editMode = .inactive
   }
 
-  func importIntoLibrary(_ items: [String]) {
+  func importIntoLibrary(_ items: [PathUuidPair]) {
     do {
-      try libraryService.moveItems(items, inside: nil)
-      syncService.scheduleMove(items: items, to: nil)
+      try libraryService.moveItems(items.map({ $0 }), inside: nil)
```
GianniCarlo: **SUGGESTION:** `items.map({ $0 })` is an identity transform — it creates an identical copy of the array. Since `items` is already `[PathUuidPair]`, this can be simplified to just `items`. Same on line 374, 397, 398, 431, 432.

Shared/CoreData/Lightweight-Models/PathUuidPair.swift
```diff
@@ -0,0 +1,19 @@
...
+//  PathUuidPair.swift
+//  BookPlayer
+//
+//  Created by Pedro Iñiguez on 6/3/26.
+//  Copyright © 2026 BookPlayer LLC. All rights reserved.
+//
+
+import Foundation
+
+public struct PathUuidPair: Equatable {
```
GianniCarlo: **SUGGESTION:** Consider also conforming to `Hashable` — this struct is frequently used in arrays and mapped. If it's ever used in a `Set` or as a dictionary key, `Hashable` conformance would be needed. Since all stored properties are `String` (which is already `Hashable`), the compiler can auto-synthesize it.

GianniCarlo (commented):
AI feedback

Shared/SwiftData/MigrationPlan.swift
```diff
@@ -0,0 +1,97 @@
...
+    },
+    didMigrate: { context in
+      guard let coreDataContext = injectedCoreDataContext else {
+        fatalError("Core Data context was not injected before migration!")
+      }
+      // 1. Fetch all V2 models (which have both path and optional uuid)
+      let items = try context.fetch(FetchDescriptor<SchemaV2.SyncTaskReferenceModel>())
+      for item in items {
+        item.uuid = "LEGACY_UUID_PLACEHOLDER"
+      }
```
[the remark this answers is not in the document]
Hirobreak: This doesn't matter since even before finishing the migration these uuids are overwritten

Shared/SwiftData/MigrationPlan.swift
Hirobreak: Added injection in ExtensionDelegate as well

Shared/SwiftData/Models/SchemaV2SyncTasksModels.swift
Hirobreak: its fine backend handles random uuids

Shared/SwiftData/MigrationPlan.swift
```diff
@@ -0,0 +1,97 @@
...
+  // Stage 2: Custom logic to populate UUIDs, then drop path
+  static let v1ToV2 = MigrationStage.custom(
+    fromVersion: SchemaV1.self,
+    toVersion: SchemaV2.self,
+    willMigrate: { _ in
+    },
+    didMigrate: { context in
+      guard let coreDataContext = injectedCoreDataContext else {
+        fatalError("Core Data context was not injected before migration!")
+      }
```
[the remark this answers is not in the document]
Hirobreak: Added injection in ExtensionDelegate as well

Shared/Services/Sync/SyncService.swift
Hirobreak: its fine, it will be very unlikely to create another task with the same conflicted uuids

GianniCarlo (approved):

PR #1491, opened 2026-03-24
Add multi-server support for Jellyfin and AudiobookShelf

matalvernaz:
## Problem

Both integrations only support a single saved server. Users with multiple Jellyfin or AudiobookShelf instances (e.g. home + work, personal + shared) have to sign out and re-enter credentials every time they want to switch.

## Solution

Store any number of server connections and expose a unified "Media Servers" experience that combines both Jellyfin and AudiobookShelf into one entry point.

**"Media Servers" (the import sheet, replaces the separate "Download from…" buttons):**
- 0 servers saved → type picker (Jellyfin / AudiobookShelf) → sign-in form
- 1+ servers saved → unified list showing all servers from both integrations with type icons; tap to browse
- "Add Server" button → type picker → connection form (handled in-place when servers of that type already exist)

**Settings screen (connection details, accessed from within each library browser):**
- Shows all saved servers with an active indicator (✓ checkmark)
- Per-server sign-out button
- "Add Server" button — shows the sign-in form; a Cancel button returns to the server list without saving

### Data model (both integrations)
- Add `id: String` (UUID) to `JellyfinConnectionData` / `AudiobookShelfConnectionData`
- Custom `Decodable` init generates a UUID for any existing saved data that lacks the field — zero-friction migration

### Connection services (both integrations)
- Keychain storage: single object → `[ConnectionData]` array
- `reloadConnections()`: tries array format first, migrates old single-object format on first launch
- `signIn()`: appends new connection, deduplicates on `url + userID`
- `activateConnection(id:)`: switch active server
- `deleteConnection(id:)`: remove specific server; `deleteConnection()` is kept for backward compatibility
- `activeConnectionID` persisted in `UserDefaults`

### New views
- `MediaServersView` — unified server list combining Jellyfin and AudiobookShelf servers, with type icons, add-server flow, and in-place add sheets
- `IntegrationServerPickerView` — per-integration server picker (used internally by root views when `skipServerPicker` is false)

### Updated views
- `ItemListView` — replaced separate "Download from Jellyfin" / "Download from AudiobookShelf" menu buttons with a single "Media Servers" button
- `MainView` — handles the new `.mediaServers` sheet case; passes `skipServerPicker: true` to root views when launched from the unified list
- `ListStateManager` — added `.mediaServers` case to `IntegrationSheet` enum
- `JellyfinRootView` / `AudiobookShelfRootView` — added `skipServerPicker` parameter to bypass the per-integration server picker when the caller has already activated the desired server
- `IntegrationConnectedView` — renders all saved servers as a list (used in Settings)
- `IntegrationConnectionView` — routes to picker vs. form based on server count and view mode; Cancel toolbar button when adding a second server from Settings

### View models (both integrations)
- `handleSignOutAction(id:)` — remove a specific server
- `handleActivateAction(id:)` — switch active server and navigate to its library
- `handleAddServerAction()` / `handleCancelAddServerAction()` — Settings-only add flow

## Testing

**No servers (fresh start):**
- Tap "Media Servers" → type picker shows Jellyfin and AudiobookShelf options
- Tap a type → sign-in form appears
- After connecting, tap "Media Servers" again → server list shows the new server

**Single server (regression):**
- Tap "Media Servers" → server list with one entry and "Add Server" button
- Tap the server → navigates directly to library, no change from before

**Multiple servers (mixed types):**
- Tap "Media Servers" → unified list shows all servers with type icons
- Tap any server → activates it and opens the correct library browser
- Tap "Add Server" → type picker → connection form → new server appears in list

**Multiple servers (same type):**
- Settings → connection details → shows all servers of that type with active checkmark
- Tap "Add Server" from unified list → pick same type → add sheet appears within the list view
- Sign in → new server appears; sign out per-server works independently

**Migration:**
- User with an existing single saved connection opens the app → connection still works; keychain is migrated to array format transparently

## Notes
- "Add Server" and "Media Servers" are new UI strings that need localization keys


GianniCarlo:
@matalvernaz after the PR that adds the different options for AudiobookShelf, I reworked how both integrations are built, and the UX for how we handle showing the connection details. I made the last change also thinking of multiple connections for the integrations, but I still have a couple of things to work on prior to the next release, I'll put all the latest changes in the beta though

matalvernaz:
Hey, thanks for the update. I see some of your changes and would be happy to keep working on this based on them. Is that something you'd want me to help with, or is there anything else that would be more useful?

GianniCarlo:
@matalvernaz hmmm I guess it's more towards what would be more useful to you, and what you want to see next in BookPlayer, multi-server support is something that I think we should do, but if you would rather do another thing, we could take a look at that too 👌

matalvernaz:
Okay awesome, I'll work on that then! I see how I'd need to change what I did. I can't think of anything else that'd be that useful yet, but will contribute when I do, and hopefully fix things to be better instead of worse.

matalvernaz:
Hey! I've rebased this onto the latest develop and added a unified "Media Servers" UI on top of the multi-server work.

**What changed:** The two separate "Download from Jellyfin" / "Download from AudiobookShelf" menu items are now replaced with a single "Media Servers" button. It opens a unified list showing all saved servers from both integrations (with type icons), and an "Add Server" button that lets you pick the server type and go through the connection flow.

I've finally managed to get hold of a macOS compilation environment, so I can test on my phone. So far I've only tested with two Jellyfin servers added, but the multi-server support seems to be working alright for me. Will continue testing with AudiobookShelf as well.

The branch is up to date with develop as of today.

matalvernaz:
Hey! Got this caught up to current develop (lots to fold in: Core Data v11, custom-headers from #1513, sticky-sort, EoC fix). Rolled in a few iterations on the multi-server UI on top.

Tapping a server now opens its library browser as a sheet on top of the unified list, with a "Media Servers" back button in the leading slot. I tried pushing via NavigationStack first, but pushing a TabView inside one auto-pops on iOS 26 — sheet-on-sheet just sidesteps that.

Also fixed a couple of dead-end exits I noticed while testing. Library picker now has an unconditional Cancel (falls back to dismissing the integration sheet if you haven't picked a library yet), and the connection-form xmark closes just the form instead of bailing out of the whole integration view. The empty state's add-server rows got restyled to match the populated server-row layout (icon + name + chevron) so they read as tappable.

Tested on TestFlight build 5.20.0 / 20260505091947 with two Jellyfin servers and one ABS. Happy to split anything out if it's easier to review separately.

GianniCarlo:
thanks @matalvernaz ! I had a mental block on getting out the current release 5.20.0 before looking into other PRs, but now that's done, I can focus again

matalvernaz:
Hey @GianniCarlo, thanks for jumping on this. Copilot caught a lot of locale-file noise, but there are three things buried in there that look like real regressions from `3c057c9` worth fixing before another pass:

1. **`AudiobookShelfConnectionService.swift:64`** — `pingServer` is an unauthenticated probe but now goes through `validateAuthenticatedResponse`, which throws `IntegrationError.sessionExpired` on 401/403 whenever any saved connection exists. The session-expired flow in `4d2a19b5` was scoped to authenticated paths on purpose, so pre-sign-in probes (e.g. `findServer`) wouldn't get mis-classified. The centralization here folds the unauth path back into the auth validator. Probably wants a 2xx-only validator for the ping path (or just keep the inline status check it had before).

2. **`JellyfinRootView.swift:608`** — `accessibilityLabel` is set to the raw key `"settings_title"` rather than `"settings_title".localized`. I'm blind and use VoiceOver, so this one actually bites me — it'll read out the key name verbatim.

3. **`BookPlayer.xcodeproj/project.pbxproj`** — duplicate UUIDs (`B14881000000000000000001` / `…0002`) for `IntegrationCustomHeadersSectionView.swift` in both `PBXBuildFile` and `PBXFileReference` sections. Looks like a leftover from folding in #1513 — strict parsers will reject it.

Happy to push the fixes myself if that's easier, just let me know. The locale-key seeds (missing `integration_add_server_button` / `media_servers_title` across ~25 `.lproj` files) I can also batch in if you want.

GianniCarlo:
Hi @matalvernaz ! thanks for pointing those out, I'll take care of them, my intend is to merge this PR today, also I have yet to respond to Alex's email, I'll get to that this weekend 👌

GianniCarlo:
it's finally merged @matalvernaz 🎉 , sorry about the delay

matalvernaz:
Hey, just wondering if you had a particular reason for separating the media servers into categories like that instead of a unified list? It might make it easier to add different types of media servers if they weren't separate. Not a judgement either way, asking to understand.

GianniCarlo:
@matalvernaz I would rather they be separate, at the keychain level the connections live in separate arrays, so they drive different lists, and also this is opinionated, but I liked better having the 'add' button per section, so the connection details is easier to reach vs a pop-up menu with the two options (jellyfin, ABS). I'm open to reworking this once more integrations  are added though

PR #1510, opened 2026-04-12
Avoid redundant @Published writes when recalculating player progress

### LINKED ISSUE #1509: UI lag during playback on lower-powered devices (iPhone SE 2)
## Description

During audio playback, the UI becomes noticeably laggy on lower-powered devices like the iPhone SE 2 (A13 chip). This affects both the player screen and the library list view.

## Root cause

Several progress-related publishers fire every ~1 second (from `AVPlayer`'s periodic time observer) without any throttling:

- **`PlayerViewModel.bindBookPlayingProgressEvents()`** — The `.bookPlaying` and `.listeningProgressChanged` notification subscribers call `recalculateProgress()` on every tick, updating ~6 `@Published` properties and triggering a full player view redraw each second.
- **`ItemProgressView`** — Both `currentProgressPublisher()` (KVO on `percentCompleted`) and `immediateProgressUpdatePublisher` fire every second for the currently playing item with no throttle, causing every visible library row's progress ring to redraw.

The existing `.folderProgressUpdated` subscriber in `ItemProgressView` already has a 1-second throttle, but the other two publishers were missed.

## Expected behavior

Smooth UI during playback on all supported devices.

## Steps to reproduce

1. Open BookPlayer on an iPhone SE 2 (or similar A13 device)
2. Start playing an audiobook
3. Navigate to the library list while playback continues
4. Observe stuttering/lag in scrolling and UI interactions

## Environment

- Device: iPhone SE 2nd generation
- iOS version: current
- BookPlayer version: latest develop

matalvernaz:
> **Maintainer edit:** this PR was refocused and rebased onto latest `develop`. Review of the original throttle approach found it couldn't reduce the 1 Hz update rate (a 500ms throttle over a 1/sec source passes every event through), and the throttle on `immediateProgressUpdatePublisher` could drop `isFinished` updates on folder rows. The branch now targets the redundant `@Published` churn directly.

## Summary

- Make `ProgressData` conform to `Equatable`
- `recalculateProgress()` now builds the new `ProgressData` locally and assigns the `@Published` property once, only when the value actually changed (previously five separate field writes, each firing `objectWillChange`)
- `hasPreviousChapter` / `hasNextChapter` are only written when their values change (previously rewritten on every playback tick)

## Context

Related to #1509. `recalculateProgress()` runs on every playback tick (1 Hz AVPlayer periodic observer), and previously fired `objectWillChange` up to seven times per tick even when most values were unchanged.

This does not fully close #1509: the library-list cost (the per-tick `lastPlayDate` fan-out to folder rows through the immediate metadata publisher, and the unthrottled `DynamicAccessibilityLabel` rebuild) is a separate follow-up.

## Test plan

- [ ] Play a chaptered book: slider, current/remaining time, chapter title, and prev/next chevrons update as before
- [ ] Toggle chapter-context and remaining-time preferences during playback
- [ ] Drag the slider across a chapter boundary and release; position sticks
- [ ] Play a book inside nested folders and verify library progress rings still update

### DIFF
--- a/BookPlayer/Library/ItemList/Views/ItemProgressView.swift
+++ b/BookPlayer/Library/ItemList/Views/ItemProgressView.swift
@@ -10,6 +10,13 @@ import BookPlayerKit
 import SwiftUI
 
 struct ItemProgressView: View {
+  /// Throttle window applied to per-row progress publishers. Each visible row in a long
+  /// library list subscribes to three high-frequency progress streams; without throttling
+  /// SwiftUI re-diffs the circular indicators on every audio tick, which lags VoiceOver
+  /// and burns CPU. One second matches the existing `.folderProgressUpdated` cadence and
+  /// keeps the list feeling live.
+  private static let progressUpdateThrottleSeconds = 1
+
   let item: SimpleLibraryItem
   let isHighlighted: Bool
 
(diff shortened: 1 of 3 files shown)


matalvernaz:
Caught up to current develop. Folded in a follow-up commit that pulls the throttle windows out into named constants with short docstrings explaining why each duration was chosen — the same literal `500` ms / `1` s shows up in a few publisher pipelines, better to have one place to tune them.

GianniCarlo:
I have to double check this, I have an iPhone SE 2 lent to someone, I just have to track it down, I'll put this on hold in the meantime

PR #1523, opened 2026-05-05
External Resource

Hirobreak:
## Purpose

- Allow for a new subscription tier to stream files from external resources (Jellyfin, Audiobookshelf)

## Related tasks

- Based on the rework of uuids for library items

## Approach

- New table **External Resource** which allows for the id, provider name and more metadata related to the source to be stored in the database

## Screenshots

IMG_8741


Shared/Services/ConcurrentSync/AsyncOperation.swift
```diff
@@ -0,0 +1,48 @@
...
+  override var isFinished: Bool { state == .finished }
+  override var isAsynchronous: Bool { true }
+  
+  override func start() {
+    guard !isCancelled else {
+      state = .finished
+      return
+    }
+    main()
+    state = .executing
```
GianniCarlo, line 40: **[MUST]** `start()` calls `main()` *before* `state = .executing`. KVO observers (including `OperationQueue`) see `isExecuting == false` while the work is already running. Any subclass that awaits inside `main()` and checks `isCancelled` mid-await is operating against stale state. Move `state = .executing` to before `main()`.

Shared/Services/ConcurrentSync/ConcurrenceService.swift
```diff
@@ -0,0 +1,261 @@
...
+    guard let nextTask = await taskContainer.getNextTask(for: queueKey) else {
+      // The queue is empty! Use scoped locking to remove the key.
+      stateLock.withLock {
+        activeQueueKeys.remove(queueKey)
+      }
+      return
+    }
+    guard let operation = createOperation(for: nextTask) else { return }
+
+    operation.onProgress = { progress in
```
GianniCarlo, line 197: **[MUST]** When `createOperation(for:)` returns `nil` (missing parameters), this `guard … else { return }` returns *without* calling `stateLock.withLock { activeQueueKeys.remove(queueKey) }`. The queue key is then permanently locked — no future `startWorkerLoop` call will unblock it, and any tasks added later for that key will silently queue forever. Mirror the empty-queue branch above and remove the key before bailing.

Shared/CoreData/Backed-Models/ExternalResource+CoreDataProperties.swift
```diff
@@ -0,0 +1,69 @@
...
+  @nonobjc public class func fetchRequest() -> NSFetchRequest<ExternalResource> {
+    return NSFetchRequest<ExternalResource>(entityName: "ExternalResource")
+  }
+
+  @NSManaged public var id: Int32
+  @NSManaged public var providerName: String
+  @NSManaged public var providerId: String
+  @NSManaged public var syncStatus: String
+  @NSManaged public var lastSyncedAt: Date?
+  @NSManaged public var processedFile: Bool
```
GianniCarlo, line 23: **[MUST]** These `@NSManaged` properties (`id: Int32`, `providerName: String`, `providerId: String`, `syncStatus: String`, `processedFile: Bool`) are non-optional in Swift but are declared `optional="YES"` in the v12 model (see `Audiobook Player 12.xcdatamodel/contents:25-31`). Any row that lands with a nil value will crash on access. Either tighten the model to `optional="NO"` with default values, or widen the Swift types to `String?` / `Int32?`. The scalar cases (`Int32`, `Bool`) are particularly subtle because Core Data silently returns 0/false for nil scalars but writes can raise constraint violations.

Shared/CoreData/BookPlayer.xcdatamodeld/Audiobook Player 12.xcdatamodel/contents
```diff
@@ -0,0 +1,107 @@
...
+        <attribute name="lastPlayDate" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="orderRank" attributeType="Integer 16" defaultValueString="0" usesScalarValueType="YES"/>
+        <attribute name="originalFileName" optional="YES" attributeType="String"/>
+        <attribute name="percentCompleted" attributeType="Double" defaultValueString="0.0" usesScalarValueType="YES"/>
+        <attribute name="relativePath" attributeType="String"/>
+        <attribute name="remoteURL" optional="YES" attributeType="URI"/>
+        <attribute name="speed" attributeType="Float" defaultValueString="1" usesScalarValueType="YES"/>
+        <attribute name="title" attributeType="String"/>
+        <attribute name="type" attributeType="Integer 16" defaultValueString="0" usesScalarValueType="YES"/>
+        <attribute name="uuid" optional="YES" attributeType="String"/>
```
GianniCarlo, line 68: **[MUST]** `uuid` regressed from non-optional in v11 to `optional="YES"` in v12, but `LibraryItem.uuid` is still declared `String` (non-optional) in Swift. Existing rows are safe (uuid was mandatory in v11 so all migrated rows have it), but every new code path must guarantee uuid is set or it'll crash on access. Either restore non-optional in the model or update Swift to `String?`.

Shared/CoreData/BookPlayer.xcdatamodeld/Audiobook Player 12.xcdatamodel/contents
```diff
@@ -0,0 +1,107 @@
...
+        <relationship name="bookmarks" optional="YES" toMany="YES" deletionRule="Cascade" destinationEntity="Bookmark" inverseName="item" inverseEntity="Bookmark"/>
+        <relationship name="externalResources" optional="YES" toMany="YES" deletionRule="Cascade" destinationEntity="ExternalResource" inverseName="libraryItem" inverseEntity="ExternalResource"/>
+        <relationship name="folder" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="Folder" inverseName="items" inverseEntity="Folder"/>
+        <relationship name="hardcoverBook" optional="YES" maxCount="1" deletionRule="Cascade" destinationEntity="HardcoverBook"/>
```
GianniCarlo: **[MUST]** `hardcoverBook` regression: in v11 this relationship had `inverseName="libraryItem" inverseEntity="HardcoverBook"` and `HardcoverBook` had a back-pointer to `LibraryItem`. In v12 both directions are gone — this relationship has no `inverseName`/`inverseEntity` and the `HardcoverBook` entity (lines 37-44) has no relationship back. Core Data warns on missing inverses and relationship-consistency guarantees are lost. Restore the inverse on both sides.

BookPlayer/Jellyfin/Library Screen/JellyfinLibraryViewModel.swift
```diff
@@ -332,6 +348,58 @@ final class JellyfinLibraryViewModel: IntegrationLibraryViewModelProtocol, BPLog
...
+        orderRank: 0,
+        parentFolder: nil,
+        originalFileName: "\(item.name).\(fileExt)",
+        lastPlayDate: item.lastPlayedDate,
+        type: .book,
+        uuid: UUID().uuidString
+      )
+      
+      let externalItem = SimpleExternalResource(
+        id: UUID().hashValue,
```
GianniCarlo: **[MUST]** `UUID().hashValue` (an `Int`) is used as the unique `id` for `SimpleExternalResource` here and at lines 588 and 812 (also `AudiobookShelfLibraryViewModel.swift:273`). Swift's `UUID` hash is not collision-free; with bulk imports the same hash can appear more than once in a single batch, producing duplicate IDs. `JellyfinAudiobookDetailsViewModel.virtualImportAudiobook:118` uses `Int(Date.timeIntervalBetween1970AndReferenceDate)` instead — that truncates to 1-second resolution and is *guaranteed* to collide for any two items processed in the same second. Use the server's own item ID (Jellyfin/ABS provider ID) or a monotonically incrementing counter.

Shared/Services/ConcurrentSync/ConcurrentSyncTask.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  ConcurrentSyncTask.swift
+//  BookPlayer
+//
+//  Created by Pedro Iñiguez on 24/3/26.
+//  Copyright © 2026 BookPlayer LLC. All rights reserved.
+//
+
+public struct ConcurrentSyncTask: Identifiable {
```
GianniCarlo, line 22: **[MUST]** `parameters: [String: Any]` is not `Sendable`. This struct crosses actor boundaries (passed from the `ConcurrentTasksRepository` actor into unstructured `Task`s in `ConcurrenceService.enqueueNextTask` and into `Operation.completionBlock`). Under Swift 6 strict concurrency this is a data-race hazard. Either make `parameters` a typed payload (e.g. an enum with associated values per `jobType`) or decode parameters into a strongly-typed struct *before* crossing the boundary.

Shared/Network/NetworkClient.swift
```diff
@@ -140,9 +140,22 @@ public class NetworkClient: NetworkClientProtocol, BPLogger {
     request.cachePolicy = .reloadIgnoringLocalCacheData
```
GianniCarlo, line 140: **[MUST]** The `upload(_:remoteURL:)` method is declared `async throws` but doesn't actually throw on non-2xx responses — failures are `print`ed and the function returns normally. The caller (`FileUploadOperation`) can't distinguish a clean upload from a server failure and the 15-second pop-then-discard path in `ConcurrenceService.enqueueNextTask` will eat the task with no user signal. Validate `(response as? HTTPURLResponse)?.statusCode` and throw a typed error on non-2xx.

BookPlayer/Library/ItemList/ItemListViewModel.swift
```diff
@@ -782,10 +790,65 @@ extension ItemListViewModel {
...
+    libraryService.handleSyncFromExternalResouce(remoteItemsDictionary: results)
+  }
 }
 
 extension ItemListViewModel: PlaybackSyncProgressDelegate {
   func waitForSyncInProgress() async {
     _ = await contentsFetchTask?.result
   }
+  
+  func fetchExternalResource(_ playableItem: PlayableItem) async {
```
GianniCarlo, line 883: **[MUST]** `fetchExternalResource` spawns an unstructured `Task { … }` with no cancellation handle and no error surface. (1) Rapid item switches race — the task continues running against a stale `playableItem`, racing against new play calls. (2) The `guard let … try await jellyfinService.fetchItem(…) else { return }` silently swallows all network errors (401, 503, expired token, server-down). Store the task as `Task?` on `self`, cancel it on the next call, and route errors to `playerState` or at least `Self.logger.error`.

BookPlayer/Player/PlayerLoaderService.swift
```diff
@@ -36,44 +36,44 @@ final class PlayerLoaderService: @unchecked Sendable {
...
-    let fileURL = DataManager.getProcessedFolderURL().appendingPathComponent(relativePath)
-
+    guard let libraryItem = self.libraryService.getSimpleItem(for: uuid)
+    else { return }
+    
+    let fileURL = DataManager.getProcessedFolderURL().appendingPathComponent(libraryItem.relativePath)
+    
     if syncService.isActive == false,
-      !FileManager.default.fileExists(atPath: fileURL.path)
+       libraryItem.externalResources?.isEmpty ?? true,
```
GianniCarlo, line 49: **[MUST]** Streaming-fallback gap: this guard says "throw if `syncService.isActive == false` AND `externalResources` is empty AND the file is missing". When a free user has an `ExternalResource` but the file is absent (the streaming case), the guard correctly skips the throw — but the downstream `getPlayableItem(from: libraryItem)` at line 73 still tries to build a `PlayableItem` from the missing local path, which will fail unhelpfully or silently. Detect `syncStatus == .stream` here explicitly and route to the streaming load path instead of relying on the guard's negative space.

BookPlayer/Import/ExternalSource/ExternalImportView.swift
```diff
@@ -0,0 +1,122 @@
...
+//
+import SwiftUI
+import BookPlayerKit
+
+struct ExternalImportView<Model: ExternalViewModelProtocol>: View {
+  @ObservedObject var viewModel: Model
+  @Environment(\.dismiss) var dismiss
+  @EnvironmentObject private var theme: ThemeViewModel
+  
+  var body: some View {
```
GianniCarlo, line 16: **[MUST]** Two issues across this view: (1) **Safe area** — the `ZStack { theme.systemBackgroundColor.ignoresSafeArea() … }` covers everything but the close/confirm buttons at lines 23-52 only get `.padding(.top, 10)`. On devices with a Dynamic Island/notch those buttons sit under the safe area. Use `.safeAreaPadding(.top)` or move dismiss into a `.toolbar`. (2) **Hardcoded `.white`** — text and icon colors throughout are `.white` / `Color.white.opacity(...)`, breaking light mode and not dark-mode-aware. Use `theme.primaryColor` / `theme.secondaryColor` like the rest of the app.

BookPlayer/Import/ExternalSource/ExternalImportView.swift
```diff
@@ -0,0 +1,122 @@
...
+          .foregroundColor(Color.white.opacity(0.6))
+          .lineSpacing(4)
+        
+        Text("\(viewModel.resources.count) File\(viewModel.resources.count == 1 ? "" : "s")")
+          .font(.headline)
+          .foregroundColor(Color.white.opacity(0.6))
+          .padding(.top, 10)
+        
+        ScrollView {
+          VStack(spacing: 0) {
```
GianniCarlo: **[MUST]** Non-lazy `VStack` inside `ScrollView` for an unbounded resource list — every row is rendered eagerly. Replace with `List` (lazy + free swipe-to-delete with `.swipeActions`, which would also remove the manual minus-button delete pattern at lines 74-82) or at minimum `LazyVStack`.

BookPlayer/Profile/CompleteAccount/SyncInvitationCard.swift
```diff
@@ -0,0 +1,85 @@
...
+            .cornerRadius(14)
+        }
+      }
+    }
+    .padding(24)
+    .background(Color(UIColor.secondarySystemGroupedBackground))
+    .cornerRadius(24)
+    .shadow(color: Color.black.opacity(0.15), radius: 20, x: 0, y: 10)
+    .overlay(alignment: .topLeading) {
+      Button(action: onCancel) {
```
GianniCarlo, line 84: **[MUST]** The `xmark.circle.fill` overlay button has no explicit frame — its tappable area is just the SF Symbol glyph (~22pt), well below the 44pt HIG minimum. Add `.frame(width: 44, height: 44)` to the button label.

BookPlayer/MainView.swift
```diff
@@ -99,10 +101,25 @@ struct MainView: View {
...
+        Text("There is another source with current play time \(TimeParser.formatTime(playerState.remotePlayTime ?? 0)). Play from this time?")
+      }
+    }
+    .sheet(isPresented: $importManager.isShowingExternalImportView) {
+      ExternalImportView(
+        viewModel: ExternalImportViewModel(importManager: importManager)
+      )
+      .presentationBackground(.clear)
+      .environmentObject(importManager)
+      .environmentObject(ThemeViewModel())
```
GianniCarlo: **[MUST]** `.environmentObject(ThemeViewModel())` injects a brand-new, unshared `ThemeViewModel` into `ExternalImportView`, so the import sheet won't reflect the user's current theme or react to system theme changes. Pass the existing `theme` `@StateObject` (already used at line 128) instead.

Shared/Services/ConcurrentSync/ConcurrenceService.swift
```diff
@@ -0,0 +1,261 @@
...
+      var params = params
+      params["id"] = UUID().uuidString
+      params["jobType"] = ExternalSyncJobType.update.rawValue
+      params["queueKey"] = queueKey
+      /// Override param `lastPlayDate` if it exists with the proper name
+      if let lastPlayDate = params.removeValue(forKey: #keyPath(LibraryItem.lastPlayDate)) {
+        params["lastPlayDateTimestamp"] = lastPlayDate
+      }
+      try await taskContainer.storeTask(parameters: params)
+    }
```
GianniCarlo, line 411: **[SHOULD]** `Task { try await taskContainer.storeTask(...) }` here (and at line 259 in `scheduleFileUpload`) is fire-and-forget with no `catch`. A SwiftData persistence error (disk full, schema mismatch) will silently disappear — the caller gets no feedback and no retry. At minimum log via `Self.logger.error`; ideally surface the error so the UI can react.

Shared/Services/ConcurrentSync/ConcurrenceService.swift
```diff
@@ -0,0 +1,261 @@
...
+  
+  func scheduleFileUpload(params: [String: Any])
+}
+
+public class ConcurrenceService: ConcurrenceServiceProtocol {
+  let operationQueue: OperationQueue
+  var taskContainer: ConcurrentTasksRepositoryProtocol! // Your DB model
+  var libraryService: LibrarySyncProtocol!
+  
+  public var accessPolicy: [ExternalSyncJobType: Bool] = [:]
```
GianniCarlo: **[SHOULD]** `accessPolicy` is a plain `public var` on a non-actor class. It's read from arbitrary Combine schedulers (`scheduleMetadataUpdate`/`scheduleFileUpload` callers) and written from `updateConcurrentService` (called on subscription-change callbacks). Unsynchronized read/write across threads = data race. Guard with `stateLock.withLock { … }` or move ownership to an actor.

Shared/Services/ConcurrentSync/ConcurrenceService.swift
```diff
@@ -0,0 +1,261 @@
...
+    }
+    
+    operation.completionBlock = { [weak self] in
+      // 2. Bridge back into the async world inside the synchronous completion block
+      Task {
+        guard let self = self else { return }
+        
+        if operation.didSucceed {
+          await self.taskContainer.pop(nextTask)
+        } else {
```
GianniCarlo, line 211: **[SHOULD]** Failure path always discards the task after one 15-second sleep regardless of error type — no retry counter, no distinction between transient errors (network blip, 503) and permanent ones (bad credentials, 404). A single transient 500 eats the task. Add a retry count on the task model and only `pop` after N attempts or on permanent errors.

Shared/Jellyfin/Network/JellyfinConnectionData.swift
```diff
@@ -52,8 +52,12 @@ struct JellyfinConnectionData: Codable {
...
 
 extension JellyfinConnectionData: CustomDebugStringConvertible {
-  var debugDescription: String {
+  public var debugDescription: String {
     let accessTokenDebugDesc = accessToken.isEmpty ? "<empty>" : "<redacted>"
     return "JellyfinConnectionData(\(url), \(serverName), \(userID), \(userName), \(accessTokenDebugDesc))"
   }
+  
+  public func buildDownloadUrl(providerId: String) -> String {
+    return "\(self.url.absoluteString)/items/\(providerId)/Download?api_key=\(self.accessToken)"
```
GianniCarlo: **[SHOULD]** Embedding `api_key=\(self.accessToken)` in the URL query string puts the credential in plaintext in server access logs, proxy logs, and (potentially) browser history if this URL leaks anywhere. Same pattern in `Shared/AudiobookShelf/Network/AudiobookShelfConnectionData.swift:61` with `token=`. Prefer the Authorization header path used elsewhere in the service layer.

Shared/AudiobookShelf/Network/AudiobookShelfConnectionData.swift
```diff
@@ -52,8 +52,12 @@ struct AudiobookShelfConnectionData: Codable {
...
 
 extension AudiobookShelfConnectionData: CustomDebugStringConvertible {
-  var debugDescription: String {
+  public var debugDescription: String {
     let apiTokenDebugDesc = apiToken.isEmpty ? "<empty>" : "<redacted>"
     return "AudiobookShelfConnectionData(\(url), \(serverName), \(userID), \(userName), \(apiTokenDebugDesc))"
   }
+  
+  public func buildAudiobookshelfDownloadUrl(providerId: String) -> String {
+    return "\(self.url.absoluteString)/api/items/\(providerId)/download?token=\(self.apiToken)"
```
GianniCarlo: **[SHOULD]** Same credential-in-URL concern as `JellyfinConnectionData.swift:61` — `token=\(self.apiToken)` ends up in access logs and proxies in plaintext. Use the Authorization header instead.

BookPlayerWatch/CoreServices.swift
```diff
@@ -44,10 +47,31 @@ class CoreServices: ObservableObject {
...
     self.hasSyncEnabled = accountService.hasSyncEnabled()
+    self.updateConcurrentService(accountService.getAccessLevel())
   }
 
   func updateSyncEnabled(_ enabled: Bool) {
     hasSyncEnabled = enabled
     syncService.isActive = enabled
   }
+  
+  func updateConcurrentService(_ accessLevel: AccessLevel) {
```
GianniCarlo, line 58: **[SHOULD]** The `switch accessLevel` → `accessPolicy` mapping (lines 58-76) is copy-pasted verbatim from `ConcurrenceService.updateConcurrentService`. When a new access level is added, both sites must be updated in lockstep. Hoist the mapping to a static helper on `ConcurrenceServiceProtocol` (or a free function) so both targets share one source of truth.

Shared/Services/ConcurrentSync/ConcurrenceService.swift
```diff
@@ -0,0 +1,261 @@
...
+  
+  public var accessPolicy: [ExternalSyncJobType: Bool] = [:]
+  // Tracks which queueKeys currently have an active worker looping
+  private var activeQueueKeys = Set<String>()
+  private let stateLock = NSLock()
+  private var disposeBag = Set<AnyCancellable>()
+  private var listeningTask: Task<Void, Never>?
+  public var tasksCountService: ConcurrentTasksCountService!
+  // Services
+
```
GianniCarlo: **[SHOULD]** `private let jellyfinService = JellyfinConnectionService()` creates a *separate* `JellyfinConnectionService` instance from the one in `AppServices`. `setup()` is called at line 57 (which calls `reloadConnection()` from keychain) so it does work in the steady state — but logout/reconnect events on the AppServices instance won't propagate to this one, and vice versa. Inject the shared `JellyfinConnectionService` into `setup(libraryService:accessLevel:)` instead of creating a private copy.

BookPlayer/Library/ItemList/ItemListViewModel.swift
```diff
@@ -782,10 +790,65 @@ extension ItemListViewModel {
...
     }
   }
+  
+  func updateFromResource(jellyfinService: JellyfinConnectionService) async {
+    // 1. Break the calculation out into a distinct local variable.
+    // (Swapping `map` to `compactMap` here to safely unwrap the `.first` optional!)
+    let jellyfinItems = self.items
+      .filter { $0.externalResources?.contains { $0.providerName == ExternalResource.ProviderName.jellyfin.rawValue } ?? false }
+    let jellyfinResources = jellyfinItems
+      .compactMap { $0.externalResources!.first(where: { $0.providerName == ExternalResource.ProviderName.jellyfin.rawValue })! }
```
GianniCarlo: **[SHOULD]** `compactMap { $0.externalResources!.first(...)! }` — the outer `compactMap` was added to drop `nil` from the outer `.first(where:)`, but the inner force-unwrap of `.first(...)!` will crash if `externalResources` is non-nil but the matching predicate doesn't hit. Drop the force-unwrap; `compactMap` already handles nils.

Shared/Integrations/IntegrationDetailsViewModelProtocol.swift
```diff
@@ -45,4 +51,6 @@ protocol IntegrationDetailsViewModelProtocol: ObservableObject {
   func fetchData()
   func cancelFetchData()
   func beginDownloadAudiobook(_ item: Item) throws
+  func virtualImportAudiobook(_ item: Item) throws
```
GianniCarlo: **[SHOULD]** `func virtualImportAudiobook(_ item: Item) throws` is declared `throws` but neither `JellyfinAudiobookDetailsViewModel` nor `AudiobookShelfAudiobookDetailsViewModel` actually throw. Callers must add needless `try` or `try?`, and any future thrown error from a non-throwing implementer is silently lost. Either drop `throws` from the protocol or add real error propagation in implementers.

BookPlayer/Import/ExternalSource/ExternalImportViewModel.swift
```diff
@@ -0,0 +1,44 @@
...
+  func removeResource(withId id: String)
+  func handleImportResources() async
+}
+
+class ExternalImportViewModel: ExternalViewModelProtocol {
+  let importManager: ImportManager
+  
+  var resources: [SimpleExternalResource] {
+    get {
+      print(importManager.externalFiles.description)
```
GianniCarlo: **[SHOULD]** `print(importManager.externalFiles.description)` runs on every SwiftUI body access to `resources`. Remove before shipping (or replace with `Self.logger.debug` if you want it for diagnostics).

BookPlayer/Profile/CompleteAccount/SyncInvitationCard.swift
```diff
@@ -0,0 +1,85 @@
...
+        }
+      }
+        
+      Text("You're about to \(subscription == .lite ? "import" : "download") \(totalItems) items. \((subscription == .pro || subscription == .lite) ? "" : "Instead of taking up space, stream them directly from your server and keep your progress synced.")")
+          .font(.subheadline)
+          .foregroundColor(.secondary)
+          .multilineTextAlignment(.center)
+      
+      VStack(spacing: 12) {
+        if !(subscription == .pro || subscription == .lite) {
```
GianniCarlo, line 46: **[SHOULD]** The "Learn more" / "Download Locally" / "Import" buttons have no `accessibilityLabel` or `accessibilityHint`. VoiceOver will read raw button text without surrounding card context (e.g. "Import" on its own is ambiguous). Add `.accessibilityHint("…")` describing what each action does.

BookPlayer/Profile/Profile/Views/ConcurrentTasksCardView.swift
```diff
@@ -0,0 +1,104 @@
...
+//  Created by Pedro Iñiguez on 31/3/26.
+//  Copyright © 2026 BookPlayer LLC. All rights reserved.
+//
+
+import SwiftUI
+import BookPlayerKit
+import SwiftData
+
+struct ConcurrentTasksCardView: View {
+  var monitor = ConcurrentTaskProgressMonitor.shared
```
GianniCarlo: **[SHOULD]** `var monitor = ConcurrentTaskProgressMonitor.shared` is a plain stored property. If `ConcurrentTaskProgressMonitor` is an `ObservableObject`, this needs `@ObservedObject var monitor` (or `@StateObject` if this view owns it) for progress updates to drive view redraws. As written, the monitor's `@Published` changes won't refresh the card.

BookPlayer/Profile/Profile/Views/ConcurrentTasksCardView.swift
```diff
@@ -0,0 +1,104 @@
...
+      Image(systemName: "chevron.right")
+        .foregroundStyle(theme.primaryColor)
+        .padding(.leading, 16)
+    }
+    .padding(16)
+    .background(theme.tertiarySystemBackgroundColor)
+    .cornerRadius(16)
+    .shadow(color: Color.black.opacity(0.08), radius: 8, x: 0, y: 4)
+    .padding(.horizontal)
+    .onReceive(
```
GianniCarlo: **[SHOULD]** `onReceive` on `concurrenceService.observeConcurrentTasksCount()` doesn't `.dropFirst()`, so the publisher fires immediately on subscription. Combined with `reloadQueuedJobs()` in `onAppear` (line 73), this triggers a double fetch on first appear. `SyncTasksCardView` already does the right thing with `.dropFirst()`; mirror that here.

BookPlayer/Player/Views/DynamicWaveLoadingView.swift
```diff
@@ -0,0 +1,64 @@
...
+          Capsule()
+            .fill(Color.white)
+          // The height is a base of 40px, multiplied by our random percentages
+            .frame(width: 4, height: 40 * waveHeights[index])
+        }
+      }
+      .opacity(isPulsing ? 1.0 : 0.0) // Fades completely out
+    }
+    .task {
+      // This loop runs continuously as long as the view exists
```
GianniCarlo: **[SHOULD]** The `while !Task.isCancelled` loop uses `try? await Task.sleep(...)`. `try?` discards `CancellationError`, so the loop won't exit promptly on view disappear (it'll re-enter and only break on the next `Task.isCancelled` check). Use plain `try await Task.sleep(...)` — `.task` cancels automatically on disappear and the thrown error exits the loop cleanly.

BookPlayer/Jellyfin/Library Screen/GridLayout/JellyfinLibraryGridView.swift
```diff
@@ -0,0 +1,178 @@
...
+                  viewModel.navigation.path.append(JellyfinLibraryLevelData.details(data: item))
+                case .userView, .folder:
+                  viewModel.navigation.path.append(JellyfinLibraryLevelData.folder(data: item))
+                }
+              }
+            }
+            .onAppear {
+              viewModel.fetchMoreItemsIfNeeded(currentItem: item)
+            }
+            .frame(
```
GianniCarlo, line 55: **[SHOULD]** `adjustSize(itemMinSizeBase, ...)` and `adjustSize(itemMaxSizeBase, ...)` recompute on every geometry change *and* duplicate the computation already done at lines 32-34 for `AdaptiveVGrid`. Cache them as `let` values at the top of the `GeometryReader` closure once, then reuse.

BookPlayer/Jellyfin/Library Screen/GridLayout/JellyfinLibraryGridView.swift
```diff
@@ -0,0 +1,178 @@
...
+//  Copyright © 2025 BookPlayer LLC. All rights reserved.
+//
+
+import BookPlayerKit
+import SwiftUI
+
+struct JellyfinLibraryGridView<Model: JellyfinLibraryViewModelProtocol>: View {
+  @ObservedObject var viewModel: Model
+
+  @ScaledMetric var accessabilityScale: CGFloat = 1
```
GianniCarlo, line 15: **[SHOULD]** Typo: `accessabilityScale` → `accessibilityScale` in this `@ScaledMetric` property. Will surface in autocomplete and future searches.

BookPlayer/Profile/Profile/Views/QueuedTasksView.swift
```diff
@@ -0,0 +1,48 @@
...
+
+import SwiftUI
+import BookPlayerKit
+import SwiftData
+
+struct QueuedTasksView: View {
+  @Environment(\.accountService) private var accountService
+  @EnvironmentObject private var theme: ThemeViewModel
+  
+  var body: some View {
```
GianniCarlo, line 46: **[SHOULD]** Two `NavigationLink` cards inside a `VStack(spacing: 0)` with a trailing `Spacer`. Since this view is a navigation destination pushed inside a `NavigationStack`, a `List` with two rows (or `Form`) would give standard inset grouping, separators, and scroll behavior for free — and matches HIG for settings-style navigation destinations.

Shared/Jellyfin/JellyfinAudiobookDetailsData.swift
```diff
@@ -0,0 +1,70 @@
...
+    tags: [String]? = nil
+  ) {
+    self.artist = artist
+    self.filePath = filePath
+    self.fileSize = fileSize
+    self.overview = overview
+    self.runtimeInSeconds = runtimeInSeconds
+    self.genres = genres
+    self.tags = tags
+    self.narrator = nil
```
GianniCarlo, line 59: **[SHOULD]** `self.narrator = nil` is hardcoded and there's no `init?(apiItem:)` path that ever sets a real value. The protocol's default `var narrator: String? { nil }` makes this compile, but it means Jellyfin items will never display a narrator in the details view, even when the API returns one. Pull narrator from the API item alongside `author`.

Shared/Services/ConcurrentSync/ConcurrenceService.swift
```diff
@@ -0,0 +1,261 @@
...
+  let operationQueue: OperationQueue
+  var taskContainer: ConcurrentTasksRepositoryProtocol! // Your DB model
+  var libraryService: LibrarySyncProtocol!
+  
+  public var accessPolicy: [ExternalSyncJobType: Bool] = [:]
+  // Tracks which queueKeys currently have an active worker looping
+  private var activeQueueKeys = Set<String>()
+  private let stateLock = NSLock()
+  private var disposeBag = Set<AnyCancellable>()
+  private var listeningTask: Task<Void, Never>?
```
GianniCarlo, line 74: **[NICE]** `listeningTask` is never cancelled — there's no `deinit` on `ConcurrenceService`. In practice the service is app-lifetime so this doesn't leak, but for correctness (and to keep tests / hypothetical re-init clean) add `deinit { listeningTask?.cancel() }`.

BookPlayer/Player/PlayerManager.swift
```diff
@@ -251,10 +282,16 @@ final class PlayerManager: NSObject, PlayerManagerProtocol, ObservableObject {
...
 
-    if syncService.isActive,
-      !FileManager.default.fileExists(atPath: fileURL.path)
+    if !FileManager.default.fileExists(atPath: fileURL.path)
     {
-      asset = try await loadRemoteURLAsset(for: chapter, forceRefresh: forceRefreshURL)
+      if syncService.isActive,
+         (forceRefreshURL || chapter.externalUrl == nil) {
+        asset = try await loadRemoteURLAsset(for: chapter, forceRefresh: forceRefreshURL)
+      } else if let externalUrl = chapter.externalUrl {
```
GianniCarlo, line 303: **[NICE]** When `chapter.externalUrl != nil` and the local file is absent, `loadPlayerItem` skips `asset.loadValues(forKeys:)` and chapter loading. The TODO comment notes `loadValues` is the bottleneck, so skipping is intentional — but `chapter.duration` is then 0 at `loadChapterOperation` time, so the guard there fails and `.bookReady` is posted with `loaded: false`. Consider loading just `"duration"` and `"playable"` async for the external-URL case.

BookPlayer/Jellyfin/Library Screen/JellyfinLibraryViewModel.swift
```diff
@@ -59,11 +54,14 @@ final class JellyfinLibraryViewModel: IntegrationLibraryViewModelProtocol, BPLog
...
   var onTransition: BPTransition<Routes>?
 
   let folderID: String?
   let recursive: Bool
+  var importManager: ImportManager?
   let connectionService: JellyfinConnectionService
+  var accountService: AccountService
   private let singleFileDownloadService: SingleFileDownloadService
 
   private var fetchTask: Task<(), any Error>?
```
GianniCarlo, line 69: **[NICE]** Re: `sortBy.didSet` at line 37 — it resets `items` and calls `fetchFolderItems` without first cancelling this `fetchTask`. The dedup guard (`fetchTask == nil`) is bypassed by the direct call from `didSet`, and the old task can still append to `self.items` after the reset. Cancel `fetchTask` before mutating state in `didSet`.

BookPlayer/Import/ImportManager.swift
```diff
@@ -150,4 +154,29 @@ final class ImportManager: ObservableObject {
...
+    }
+    
+    if self.libraryService.findResource(for: externalResource.providerId, context: nil) != nil {
+      return true
+    }
+    
+    return false
+  }
+  
+  func processExternalFiles() {
```
GianniCarlo, line 175: **[NICE]** `processExternalFiles` mutates `self.externalFiles` (a `@Published` property) and sends on `externalOperationPublisher` — both should happen on `@MainActor`. Caller (`ExternalImportViewModel.handleImportResources`) is `async` with no actor constraint. Mark `processExternalFiles` as `@MainActor` or ensure call sites are.

BookPlayer/Profile/Login/ExternalSyncInfoView.swift
```diff
@@ -0,0 +1,110 @@
...
+        LoginBenefitSectionView(
+          imageName: "arrow.triangle.2.circlepath",
+          title: "Two-Way Progress Sync",
+          subtitle: "Your listening progress stays perfectly synced with your server and across all your devices."
+        )
+        
+        LoginDisclaimerSectionView()
+      }
+      .applyListStyle(with: theme, background: theme.systemGroupedBackgroundColor)
+      .safeAreaInset(edge: .bottom) {
```
GianniCarlo, line 50: **[NICE]** `Color.clear.frame(height: 88)` is a magic number compensating for the overlaid bottom CTA. The `.safeAreaInset(edge: .bottom)` approach is correct but the 88pt height should be derived from a layout preference or `GeometryReader` on the CTA itself rather than hardcoded — otherwise text-size changes will break the spacing.

BookPlayer/Profile/Profile/Views/SubscriptionPillView.swift
```diff
@@ -0,0 +1,29 @@
...
+
+import SwiftUI
+
+struct SubscriptionPillView: View {
+  let title: String
+  let backgroundColor: Color
+  let foregroundColor: Color
+  
+  var body: some View {
+    HStack(spacing: 4) {
```
GianniCarlo: **[NICE]** `HStack(spacing: 4)` wraps a single `Text`. The `HStack` is redundant — make the `Text` the direct content of the pill's padding modifiers.

BookPlayer/MediaServerIntegration/Library Screen/Details/IntegrationAudiobookDetailsView.swift
```diff
@@ -148,4 +143,75 @@ struct IntegrationAudiobookDetailsView<
     }
     .scrollIndicators(.hidden)
   }
+  
+  var DownloadButton: some View {
```
GianniCarlo: **[NICE]** `DownloadButton`, `SmallDownloadButton`, and `SynchronizeButton` are computed `var` properties returning `some View`. SwiftUI rebuilds all three every body eval even though at most two are visible. Extract them as `private` `@ViewBuilder` methods or dedicated `View` structs to let SwiftUI skip identity diffing on hidden branches.

Shared/CoreData/Lightweight-Models/SimpleExternalResource.swift
```diff
@@ -0,0 +1,55 @@
...
+
+import Foundation
+
+public struct SimpleExternalResource: Identifiable {
+  public let id: Int
+  public let providerName: String
+  public let providerId: String
+  public let syncStatus: String
+  public var lastSyncedAt: Date?
+  public var processedFile = true
```
GianniCarlo: **[NICE]** Struct default `processedFile = true` but `ExternalResource.create(_:libraryItem:)` always writes `entity.processedFile = false`. Reads back through `SimpleExternalResource(from: item)` reflect the stored value (false), not the struct default. Not a bug if `processedFile` is only ever read from DB-backed objects, but the divergence is confusing. Align them or document.

GianniCarlo (commented):
AI feedback:

BookPlayer/Player/PlayerLoaderService.swift
Hirobreak: this is fine since it builds the url path based on whether it is external or not

Shared/Services/ConcurrentSync/ConcurrenceService.swift
Hirobreak: I want to avoid too many injections, this is fine, task would just retry

Shared/Jellyfin/JellyfinAudiobookDetailsData.swift
Hirobreak: but Audiobookshelf does

GianniCarlo:
Thanks @Hirobreak for this work! develop moved ~70 commits since this branch's base and the fork has maintainer edits disabled, so we've rebased it ourselves and continued in #1586 — your commits are preserved as the feature commit's author there. Closing this one in favor of #1586.

PR #1524, opened 2026-05-05
Don't crash auto-play when the audio session is held by another process

matalvernaz:
`PlayerManager.play(autoPlayed:)` calls `fatalError()` if either `setCategory` or `setActive(true)` throws. Background auto-play — CarPlay, headphones, lock-screen, scheduled — routinely hits states where another process owns the audio session (phone call, navigation, voice memo) or the OS denies activation, and the app hard-crashes with `EXC_BREAKPOINT` from the fatalError trap. Saw this on TestFlight build 5.19.0 / 20260428053421 (iPhone 15 Pro / iOS 26.5, role=Background, `PlayerManager.swift:876`).

Two commits:

1. **Stop crashing.** Replace the fatalError with a log-and-return: NSLog the underlying error, clear the queued-playback flag, bail. A missed auto-play is strictly better than the app dying in the background.

2. **Recover when the contention clears.** On failure, leave `playbackQueued = true` so the existing `handleAudioInterruptions` path retries `play(autoPlayed:)` once iOS posts the `.ended` notification (call ended, navigation done). Also bind the interrupt observer up front so we still hear `.ended` even if activation fails. Previously a contested auto-play was a dead-end until the user manually pressed play.

### DIFF
--- a/BookPlayer/Player/PlayerManager.swift
+++ b/BookPlayer/Player/PlayerManager.swift
@@ -39,6 +39,19 @@ final class PlayerManager: NSObject, PlayerManagerProtocol, ObservableObject {
   @Published private var isFetchingRemoteURL: Bool?
   /// Prevent loop from automatic URL refreshes
   private var canFetchRemoteURL = true
+  /// Pending audio-session retry task scheduled when activation fails (beta builds only).
+  private var audioSessionRetryTask: Task<Void, Never>?
+  /// Beta builds (TestFlight or DEBUG) attempt to recover from audio-session
+  /// activation failures instead of crashing. Release builds keep the original
+  /// fatalError so we don't mask the bug in the App Store version until the
+  /// recovery path has proven itself in the field.
+  private static let isBetaBuild: Bool = {
+    #if DEBUG
+    return true
+    #else
+    return Bundle.main.appStoreReceiptURL?.lastPathComponent == "sandboxReceipt"
+    #endif
+  }()
   private var hasObserverRegistered = false
   private var observeStatus: Bool = false {
     didSet {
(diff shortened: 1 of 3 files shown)


GianniCarlo:
the crash is by design, it's hard to reproduce, but I've reached the place where the app just can't activate the audio session, and the only way to get it to work again is to force close it. I filed a radar some years ago, but nothing came of it. If we can put it so it's only for beta builds, I'm fine with it, just to see if we get reports again that the play button is not working

matalvernaz:
Thanks for the context — that's a fair concern. I dug into the recovery path and you're right that on its own it wouldn't have rescued the state you're describing: the `.ended` interrupt notification only fires for sessions that were already active and got interrupted, so a `setActive(true)` failure had nothing to "end" and the UI would sit queued forever.

Reworked the change to address both points:

- **Beta-gated**, as you asked. Release builds keep the original `fatalError` so the App Store version isn't masked while this is unproven; the new path only runs on TestFlight + DEBUG (`Bundle.main.appStoreReceiptURL?.lastPathComponent == "sandboxReceipt"`).
- **Bounded retry on activation failure**: 1s / 3s / 8s backoff, each attempt deactivates with `.notifyOthersOnDeactivation` then re-activates — the release/re-grab pattern that's known to unstick the contested-session state. If any retry succeeds, playback resumes; if all three fail, `playbackQueued` is cleared and an alert tells the user to close and reopen the app, which gives them the same actionable hint the crash used to.
- The retry task is cancelled from `play`, `pause`, and `stopPlayback`, so user actions aren't clobbered by a late retry firing.

If beta reports still come in saying the play button is unresponsive, we'll have a clear telemetry signal — either the alert is showing (recovery is failing cleanly), or we never reach it (the activation throw isn't even happening and something else is wrong). Either way the data should be more useful than a silent hang.

PR #1579, opened 2026-08-20
Fix progress label sheet animation

john-flanagan:
## Bugfix

The progress label jittered strangely when moving the player overlay.

## Approach

Scope disabling animations on the progress label to only text updates.

## Screenshots

| Before | After |
| -------- | -------- |
|  <video src="https://github.com/user-attachments/assets/152f21eb-55a1-4329-b03c-2f3a07516857" controls preload></video> |  <video src="https://github.com/user-attachments/assets/201b26cf-cde1-4682-8031-00a537827555" controls preload></video> |


GianniCarlo:
Thanks for this @john-flanagan ! I'll include it in the next release

PR #1587, opened 2026-08-26
Refresh most-recent sort on book playback

sashplatonov:
## Purpose

When the library is sorted by **Most recent**, the order comes directly from `lastPlayDate` in the fetch request.

Starting playback updates `lastPlayDate`, but the visible list is not fetched again. Because of that, the book the user just started does not move to the top until the screen is reopened or manually refreshed.

This change refreshes the **Most recent** list when playback starts, while keeping the current scroll position

## Approach

* `ItemListView.swift`

  * Listen for `.bookPlayed`.
  * Refresh only when the current location is sorted by `mostRecent`.
  * Skip the refresh while searching or editing.

* `ItemListViewModel.swift`

  * Add `reloadItemsPreservingOffset()`.
  * It reloads the currently visible range instead of resetting the list to the first page, so long lists keep their scroll position.

* `ItemListViewModelReloadTests.swift`

  * Add coverage for empty and unscrolled lists, scrolled lists, re-sorting, exhausted pagination, and the difference from the existing `reloadItems()` behavior.

`.bookPlayed` is used intentionally instead of `.bookPlaying` or playback progress updates. `lastPlayDate` is updated every second during playback, so reacting to those events would constantly re-fetch and reorder the list. Refreshing once when playback starts is enough.

* `reloadItemsPreservingOffset()` currently sets `canLoadMore` based on whether the fetch returned exactly the requested limit. At a page boundary, a re-sort could theoretically make the returned window smaller and stop pagination early.
* `.bookPlayed` also fires when autoplay starts the next book. This is expected and keeps that book near the top as well.
* The subscription stays active while another tab is selected, so a background refresh can happen before the user returns to the list.
* Query-time sorting by `lastPlayDate` is already covered by `testMostRecentReflectsPlaybackImmediately`. The new tests focus on the ViewModel reload behavior and pagination state.

### DIFF
--- a/BookPlayer.xcodeproj/project.pbxproj
+++ b/BookPlayer.xcodeproj/project.pbxproj
@@ -293,6 +293,7 @@
 		5163037679ED3EE0D1866A8F /* AppServices.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0FC8B923A5506F865253B97C /* AppServices.swift */; };
 		52AA441379FD94339FEECB55 /* SortPreferencesResolving.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5C6BC9E7E76B9BAB7A0D88D2 /* SortPreferencesResolving.swift */; };
 		59384C1A0B33B9EB42AE4C75 /* ChaptersViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F496A2BF4CBFFC745C453D42 /* ChaptersViewModelTests.swift */; };
+		A1B2C3D4E5F60001LISTV01 /* ItemListViewModelReloadTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A1B2C3D4E5F60002LISTV01 /* ItemListViewModelReloadTests.swift */; };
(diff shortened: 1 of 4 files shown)


GianniCarlo:
Hi @sashplatonov ! I saw this issue and I think it was solved in this PR ( https://github.com/TortugaPower/BookPlayer/pull/1582 ), the problem is that AppStore reviews are taking too long now, so v5.22.0 is stuck in waiting for review, I'll revisit this PR once v5.22.0 is out

GianniCarlo:
following up on this @sashplatonov , can you double check on v5.22.0 which is now out on the AppStore and see if the issue still happens?

sashplatonov:
@GianniCarlo 
I checked again on v5.22.0 from the App Store, it's **much better now, thanks.**

Sorting from #1582 works as expected: choosing "Most recent", "Title", or "File name" updates the order immediately and keeps it after restart.

There’s just one thing left:
With Most recent, starting a book updates lastPlayDate, but the list itself doesn’t refresh. So a book started from the middle stays there until you reopen the screen or restart the app.

This PR fixes exactly that: it refreshes the list once when playback starts, while keeping the current scroll position.

So with v5.22.0 + this PR, **Most recent becomes fully live, if that makes sense to you**

PR #1596, opened 2026-09-08
Replace the reviewer harness with the hardened one

GianniCarlo:
This repo still runs the first-generation reviewer: `permissionMode: 'bypassPermissions'` with unrestricted Bash, `gh pr diff` executed by the agent with a GitHub token in its environment, nothing redacted on the way out, `npm install` with the SDK at `latest` and no lockfile, and no tests.

After this PR the harness is the same code as [bookplayer-android#114](https://github.com/TortugaPower/bookplayer-android/pull/114), [bookplayer-api#39](https://github.com/TortugaPower/bookplayer-api/pull/39) and [bookplayer-support-pipeline#15](https://github.com/TortugaPower/bookplayer-support-pipeline/pull/15), differing only in the prompt, the redaction patterns and the test fixtures.

### What the agent can do now

| | Before | After |
| --- | --- | --- |
| Permission model | `bypassPermissions` | `canUseTool` gate, `permissionMode: 'default'`, `settingSources: []`, explicit `tools` list |
| Bash | anything | allowlisted read-only commands; no interpreters, test runners, `gh`, `curl`, redirects (`<` and `>`), `$`-expansion, backticks, substitution, unquoted braces, `cd`, or symlink-dereference flags |
| Paths | anywhere on the runner | confined to the checkout and the diff file, symlinks resolved, `..` refused; `~`, `/proc`, `.git/config`, `.ssh`, `.npmrc`, `.netrc`, `.env` denied |
| Environment | the whole job env, `GH_TOKEN` included | filtered by pattern: no token, secret, password, key or webhook; `GH_TOKEN`/`GH_REPO` removed from the job |
| Diff | `gh pr diff` run by the agent | fetched over REST by the harness, written to a file the agent may read (with a per-file fallback for PRs GitHub refuses to render) |
| Install | `npm install`, SDK at `latest` | `npm ci --ignore-scripts` against a committed lockfile, SDK at `^0.3.261` |
| Checkout | default | `persist-credentials: false`, so the token is not left in `.git/config` |
| Output | posted as-is | `redact()`: Sentry DSN, RevenueCat key, PEM signing keys, bearer JWTs |

### What the review itself gains

- Findings are fingerprinted: one survives a push instead of being re-posted, and stale ones resolve.
- A second pass verifies every still-open finding against the current commit and reports fixed / still open / no longer applies, instead of inferring resolution from silence. Your replies on a thread inform that judgement; they cannot close it on their own unless someone other than the PR author says so.
- The model is resolved at runtime — newest Opus tier, Opus 5 today — instead of pinned to `claude-opus-4-8`, with an ordered fallback list so a retired id cannot take the reviewer offline.
- **The second model call that repaired broken JSON is gone.** Extraction is deterministic now: position-aware fenced-block parsing with a bounded truncation repair. A repaired or limit-cut answer is still posted, but marked provisional and resolves nothing.
- A 14-minute wall-clock deadline degrades to a visible note rather than a cancelled job that may have half-reconciled the PR, and every failure path explains itself on the PR instead of leaving a red check with no comment.

### Notes for this repo

- The runner stays `ubuntu-latest`: the reviewer reads the diff and the source, it never builds the app.
- The prompt keeps this repo's emphases (retain cycles and `[weak self]` in closures and Combine sinks, stored cancellables, `@MainActor` and thread-correct database access, the player and AVAudioSession lifecycle, the `BookPlayerKit` boundary for the watch app, widgets and extensions) and adds the rule that a bare SwiftUI string literal localises itself, so only a computed `String` is a localisation bug.
- `review-guide.md` changes by one step: read the diff file instead of running `gh pr diff`. The rubric is untouched.
- 70 harness tests (`node --test`) gate the review step. They cover the allowlist, including the escapes found and closed during the Android port (brace expansion, `-f/etc/passwd`, `grep`'s pattern-vs-path ambiguity, stdin redirection), the redaction patterns, JSON extraction against real model output, and the reconcile/verify trust rules.

### DIFF
--- a/.github/claude/review-guide.md
+++ b/.github/claude/review-guide.md
@@ -11,7 +11,8 @@ auth/entitlement** bugs are the highest-priority findings.
 
 ## How to review
 
-1. Get the diff: `gh pr diff <number>`. The branch is checked out in the working directory.
+1. Read the unified diff the harness wrote for you; its path is in the task prompt. The PR branch is
+   checked out in the working directory.
 2. **Do not review the diff in isolation.** For each non-trivial change, open the surrounding code and its
    callers with `Read`/`Grep`/`Glob` before judging. Diff-only opinions are not acceptable.
 3. Respect the layout: app source is under nested `BookPlayer/BookPlayer/`; shared/framework code is in
@@ -24,7 +25,7 @@ auth/entitlement** bugs are the highest-priority findings.
 - `BookPlayer/Generated/` (Sourcery output, incl. `AutoMockable.generated.swift`), `Assets.xcassets`,
   `build/`, and any generated files.
 - `.lproj` translation files: flag a **missing `Base.lproj/Localizable.strings` key**, but do not nitpick the
-  wording of existing translations.
+  wording of existing translations (Lokalise-managed).
 - `project.pbxproj` merge noise — only comment on a genuine structural problem (wrong target membership,
   broken build setting), not diff churn.
 - **Style the tools already own — do NOT flag:** `force_try` / `try!` in general, long lines, short/`identifier_name`,
--- a/.github/claude/review-guide.md
+++ b/.github/claude/review-guide.md
@@ -7,12 +7,13 @@ concurrency, and secrets conventions. Judge changes against it. This app handles
 per-user cloud sync, auth, and subscriptions**, so **memory/concurrency, player lifecycle, threading, and
 auth/entitlement** bugs are the highest-priority findings.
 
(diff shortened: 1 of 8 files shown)


GianniCarlo:
Closing: this port ran ahead of the Android PR it is copied from, which has not converged yet, and it should have been asked for before it was opened on this repo. Will re-port from the merged Android harness, with the repo-specific findings from these rounds folded in (the Sentry DSN stored without its scheme in xcconfig, the base-branch default, the guide pointing at origin/develop on a depth-1 checkout).

